PR #5, opened 2018-06-12
fix information typo and regenerate css files

hwigmore:
@emma-floored i just changed the variables in `text.styl` and then ran `gulp dev:css`. i was trying it from within the workplace repo but i think since i originally did `git clone` instead of `git submodule` it just...became a branch / commit directly on blocks?


emma-floored:
👍

PR #6, opened 2018-06-18
add util file for generic helper styles

hwigmore:
i noticed that the `hidden` class (used in Dropdown and presumably elsewhere) wasn't in `blocks.css` yet

i'm not sure if you had a different way that you were planning to include these types of classes, but i went with a `utilities.styl` file. happy to rename / move if you prefer.


emma-floored:
oops thanks @hwigmore... I put those types of things in a `state.styl` file that I then forgot to include here, except in the docs css. Looking over it now, though, most of them _are_ DataVis specific.

Would you mind adding this one as well?
```
.disabled {
  opacity: 0.5;
  pointer-events: none;
}
```

hwigmore:
sure thing. state is a nice name. i'll change!

hwigmore:
updated!

emma-floored:
👍

PR #7, opened 2018-06-20
tweak button line height to vertically center text

hwigmore:
@emma-floored this might not be the most precise but i think it does help (especially on the small button)

master:
screen shot 2018-06-20 at 2 41 16 pm

branch:
screen shot 2018-06-20 at 2 44 25 pm


emma-floored:
💐 👍 

this has been driving me nuts, but I never did anything about it b/c the font might change 😬 but who knows how long it will take to settle on one so this definitely makes sense for now

PR #8, opened 2018-06-20
remove deprecated React.createClass

hwigmore:
@emma-floored even though i fixed this on spacer by fixing that webpack thing, i figured i should also fix the bug here since `React.createClass` doesnt work with blocks' version of react. 

a few other changes came with changing `createClass` -> `extends React.Component`, will note them below

i tested this in Spacer by including a blocks dropdown component and removing my absolute path fix


react/wrappers/closeOnClick.jsx
```diff
@@ -1,61 +1,78 @@
...
-  return React.createClass({
-    getInitialState: function() {
-      return {
-        isOpen: !!this.props.isOpen
-      };
-    },
-    componentDidMount: function() {
-      // TODO: not best practice to use findDOMNode
-      // see if we can use refs instead
-      this.domElement = ReactDOM.findDOMNode(this);
```
hwigmore: this was causing errors to be thrown. i just went ahead and wrapped `WrappedComponent` in a div below to get the ref. not as cute but it works.

react/wrappers/closeOnClick.jsx
```diff
@@ -1,61 +1,78 @@
...
+
+    componentWillUnmount() {
+      document.body.removeEventListener('click', this.closeOnClick);
+    }
+
+    open() {
       this.setState({
         isOpen: true
-      }, () => {
-        document.body.addEventListener('click', this.closeOnClick);
```
hwigmore: i just simplified things a bit by adding the event listener on mount and removing it on unmount, instead of adding and removing anytime the dropdown is toggled and only removing on unmount if it's open

emma-floored: will that work for multiple dropdowns?

hwigmore: seems to be work fine! 

- each dropdown opens and closes appropriately using its own dropdown trigger
- clicking outside of either dropdown causes whichever is open oto close
- if one dropdown is open and i click the other dropdown's trigger, the first one closes and the second opens

emma-floored: I think that's what's desired - that is how a standard combo box dropdown would work, anyway.

emma-floored:
👍 sweet, thanks @hwigmore!

PR #10, opened 2018-06-27
update blocks docs with some dropdown menu examples, fix a few things…

emma-floored:
… with the dropdowns

@hwigmore this adds a documentation page for dropdowns (they're not like, interactive b/c of JavaScript and things, but you can see the structure and what they look like)

Also fixed a few issues with dropdowns:
- the way the menus wiggle on hover b/c of the border width change
- got rid of some weird margins
- scooched up the arrow (it looked a little awkward in it's active state, like it was too low)
- some hover states on menu items weren't quite right

screen shot 2018-06-27 at 5 16 24 pm

### DIFF
--- a/docs/components/dropdowns.html
+++ b/docs/components/dropdowns.html
@@ -154,7 +154,7 @@ definition: An input that allows users to make a single selection from a pre-def
   <div class="l-flex-vertical">
     <div class="l-flex-horizontal">
       <span class="doc-state-title">
-        Active
+        Default
       </span>
       <div class="doc-state-content">
         {{ dropdown_trigger_example }}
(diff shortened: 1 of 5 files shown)


docs/components/dropdowns.html
```diff
@@ -7,13 +7,189 @@
...
+  </div>
 {% endcapture %}
 
 <!-- state & behavior section -->
 <div class="l-flex-vertical doc-section">
   <span class="heading-medium">States & Behavior</span>
   <div class="l-flex-vertical">
+    <div class="l-flex-horizontal">
+      <span class="doc-state-title">
+        Active
```
hwigmore: i think we wanna call this one Default

hwigmore:
yuck re: that one px shift on hover, but nice solution!

👍

PR #12, opened 2018-07-03
text field updates

hwigmore:
- update `TextField.jsx` so it is a pure function and its owner manages state
- add attributes to `TextField.jsx` 
- update styles so TextField can also be email or password (`.text-field input` instead of specifying input type)
- remove pointer events from readonly input

from https://github.com/floored/workplace/pull/409


hwigmore:
oh also @MagdalenaHenke i wanted to tag you on this as well / update you on how to do this: once this was merged into master i just did `git subrepo push lib/blocks -b text-field-updates` and it created the branch right here for me to open a pr!

blocks.css
```diff
@@ -1570,39 +1570,45 @@ a.disabled {
...
   border: 1px solid var(--brand-02, #75e6c9);
 }
-.text-field input[type="text"].focus {
+.text-field input.focus {
   outline: none;
   border: 1px solid #75e6c9;
   border: 1px solid var(--brand-02, #75e6c9);
 }
-.text-field input[type="text"].disabled {
+.text-field input:-moz-read-only {
```
emma-floored: do you need this one? autoprefixer should handle it I would think

emma-floored: oops n/m this is autoprefixed b/c this is the final css file, I missed the break...

emma-floored:
👍

PR #15, opened 2018-07-10
Clean up react components

emma-floored:
@hwigmore @jforzano Couple of react component changes!

- Tabs split into 2 options
  - RadioTabs: tabs implemented as radio buttons
  - LinkTabs: tabs implemented as a list of links
- RadioButtonGroup
  - previously RadioButton was just a single one, how useful is that


react/input/RadioButton.jsx
```diff
@@ -21,9 +21,9 @@ const RadioButton = props => (
 );
 
 RadioButton.propTypes = {
+  name: PropTypes.string.isRequired,
   text: PropTypes.string,
   value: PropTypes.string,
-  name: PropTypes.string,
   isDisabled: PropTypes.bool,
   isChecked: PropTypes.bool,
   onChange: PropTypes.func
```
hwigmore: not pr specific, but i feel like `onChange` should be required, no?

react/input/RadioButtonGroup.jsx
```diff
@@ -0,0 +1,43 @@
...
+      }
+    </div>
+  );
+};
+
+RadioButtonGroup.propTypes = {
+  options: PropTypes.arrayOf(PropTypes.string),
+  isDisabled: PropTypes.bool,
+  checkedIndex: PropTypes.number,
+  onChange: PropTypes.func
```
hwigmore: i think this should be required as well

react/input/RadioButtonGroup.jsx
```diff
@@ -0,0 +1,43 @@
+const React = require('react');
+const PropTypes = require('prop-types');
+
+const RadioButton = require('./RadioButton.jsx');
+
+
+const RadioButtonGroup = (props) => {
+  const name = Math.random().toString();
```
hwigmore: any reason to not make `name` a passed prop from the caller? seems like that would make sense and then stay consistent instead of changing on each render. i feel like it would be useful too in some cases for the caller to have the name attribute (like if this was part of a form and all the form data was being sent?)

react/tabs/LinkTabs.jsx
```diff
@@ -0,0 +1,42 @@
...
+    name: PropTypes.string,
+    href: PropTypes.string
+  })),
+  activeTabIdx: PropTypes.number,
+  isDisabled: PropTypes.bool,
+  isVertical: PropTypes.bool
+};
+
+LinkTabs.defaultProps = {
+  tabs: [],
```
hwigmore: can make this required and remove from `defaultProps`

react/tabs/RadioTabs.jsx
```diff
@@ -0,0 +1,49 @@
...
+  isVertical: PropTypes.bool,
+  onChange: PropTypes.func
+};
+
+RadioTabs.defaultProps = {
+  tabs: [],
+  activeTabIdx: 0,
+  isDisabled: false,
+  isVertical: false,
+  onChange: () => {}
```
hwigmore: sorry i'm being a broken record. `tabs` and `onChange` seem like required props

react/input/RadioButtonGroup.jsx
emma-floored: oh that's fair... I only did this because I always forget to give radio buttons a name and then they don't work

emma-floored: I could just make it required though

emma-floored:
@hwigmore updated!

hwigmore:
some errors to resolve on the react preview page, but otherwise 👍 
screen shot 2018-07-11 at 4 06 20 pm

PR #16, opened 2018-07-12
New font graphik

emma-floored:
@hwigmore new font plus some miscellaneous updates to things like radio buttons

screen shot 2018-07-12 at 2 16 04 pm


blocks-styles/reset.styl
```diff
@@ -48,14 +48,14 @@ textarea
...
+*
   box-sizing: border-box;
 
-*
-  box-sizing: inherit;
-  &:before,
-  &:after
-    box-sizing: inherit;
+// *
+  // box-sizing: inherit;
```
emma-floored: how important is this part to y'all? b/c it was kinda messing up some stuff... but I can try to work around it if it's necessary

hwigmore: interesting. i feel like whichever is more common should be done in blocks and then if we want something different specific applications can override in their own stylesheet

but out of curiosity i'll test it out on spacer and see if things look insane

emma-floored: for reference: https://developer.mozilla.org/en-US/docs/Web/CSS/box-sizing

(box-sizing means if you set a width/height, it includes padding and borders - default behavior is to add padding and borders to the specified width/height)

hwigmore: oh so this reset (which you are removing) would not account for border width? if so, then yes i approve! i think we want to consider border width in the sizing

emma-floored: ok sweet

blocks-styles/text.styl
```diff
@@ -1,86 +1,110 @@
...
   font-family: $sans-serif-font-family-medium;
   font-size: 96px;
+  line-height: 123px;
   color: $text-01;
 }
 
 h2 {
   font-family: $sans-serif-font-family-medium;
   font-size: 60px;
+  line-height: 77px;
```
hwigmore: these line heights are so strange and specific. i dont feel strongly but sort of want to advocate for a line-height equal to the font-size, and dealing with extra space around via padding

blocks-styles/text.styl
```diff
@@ -1,86 +1,110 @@
...
+  &.disabled {
+    color: $text-01;
+    opacity: 0.5;
+    pointer-events: none;
+  }
 }
 
 h1 {
   font-family: $sans-serif-font-family-medium;
   font-size: 96px;
```
hwigmore: def not necessary for this PR but something on my mind. i like how bulma defines a few [variables of sizes](https://github.com/floored/workplace/blob/master/assets/css/shared/utilities/initial-variables.sass#L45-L52) and uses them for consistency throughout style definitions. just something for us to think about!

blocks-styles/text.styl
emma-floored: I used these b/c they are what @runigoswami specified in the spec

hwigmore:
honestly i skimmed but i'm gonna say 👍!

blocks-styles/text.styl
emma-floored: oh I like that

hwigmore:
sorry for conflicts! 😬 😭

emma-floored:
ooooooooooooof

emma-floored:
put back in franklin gothic for now (left in comments so it should be easy to switch) - want to keep new text styles though

PR #17, opened 2018-07-13
text validation

hwigmore:
i added a bit of state in the react preview to show off the new validity functionality. might want to look at https://github.com/floored/blocks/pull/17/files?w=1 to see it without whitespace changes

also had to add state back into `TextField.jsx`

i made it so that after your initial invalid message, when you type, we check again so you dont have to do another blur to make that error go away


hwigmore:
https://app.clubhouse.io/floored/story/9025/validate-text-input-on-blur

react/input/TextField.jsx
```diff
@@ -1,88 +1,99 @@
...
-  let invalidLabelMarker = null;
-  let invalidLabelMessage = null;
-  if (!props.isValid(props.value)) {
-    invalidLabelMarker = (
-      <span className="invalid-label-marker">*</span>
-    );
-    invalidLabelMessage = (
-      <span className="invalid-label-message">{props.invalidErrorMessage}</span>
-    );
+    this.state = { isValid: true };
```
hwigmore: oh maybe this shouldn't start as `true` but should start as `isValid: props.isValid(props.value)`... hm

react/input/TextField.jsx
```diff
@@ -1,88 +1,99 @@
...
-  let suffixElement = null;
-  if (props.suffix) {
-    suffixElement = (
-      <span className="text-field-suffix">{props.suffix}</span>
-    );
+    // red asterisk if the value is invalid
+    let invalidLabelMarker = null;
+    let invalidLabelMessage = null;
+    if (!this.state.isValid) {
+      invalidLabelMarker = <span className="invalid-label-marker">*</span>;
```
hwigmore: oh i deleted a lot of extra lines just for less scrolling but this is totally personal pref and if you are against it i'll put them back!

emma-floored:
I'm curious about the changes it's showing for `package-lock.json`... do you know what's up with that?

hwigmore:
oh shoot i didnt mean to commit those. although i am curious too. whenever i `npm install` blocks (whether as a suprepo or in its own repo) i get updates there. will check it out.

react/input/TextField.jsx
```diff
@@ -1,88 +1,99 @@
...
-      <span className="invalid-label-marker">*</span>
-    );
-    invalidLabelMessage = (
-      <span className="invalid-label-message">{props.invalidErrorMessage}</span>
-    );
+    this.state = { isValid: true };
+  }
 
-    textFieldClasses += ' invalid';
+  udpateValidity() {
```
emma-floored: `update`

react/input/TextField.jsx
emma-floored: I'm fine with it, I think I mostly just have trouble processing the `>;` combination of characters... it seems crazy

react/input/TextField.jsx
```diff
@@ -1,88 +1,99 @@
...
+            this.props.onBlur(e);
+          }
+        }}
+        onKeyUp={(e) => {
+          if (e.keyCode === 13) { // ENTER
+            e.target.blur();
+          } else if (this.props.onKeyUp) {
+            this.props.onKeyUp(e);
+          }
+          // If an inValid message has already appeared via onBlur,
```
emma-floored: I know this is just a comment but the camel case here is kinda funny

react/input/TextField.jsx
MagdalenaHenke: ☝️ 👏

hwigmore:
ah yeah it was become i was on an older version of node and therefore npm. will get rid of those package-lock changes

emma-floored:
👍

PR #20, opened 2018-07-18
Add onFocus prop to TextField

hamham91:
Noticed `onFocus` was missing so I thought I would add it.

@emma-floored @hwigmore


hwigmore:
👍

emma-floored:
👍

PR #21, opened 2018-07-18
Propagate enter in onKeyUp for TextField

hamham91:
I don't know if we want exactly this, but I want to capture an `enter` in the parent component's `onKeyUp` so I changed this to always pass through the key event.

@emma-floored @hwigmore

### DIFF
--- a/react/input/TextField.jsx
+++ b/react/input/TextField.jsx
@@ -78,6 +78,7 @@ class TextField extends React.Component {
             this.updateValidity();
           }
         }}
+        onKeyDown={this.props.onKeyDown}
       />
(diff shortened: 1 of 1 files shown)


hwigmore:
yeah this seems like the behavior we'd want good call 👍

emma-floored:
👍

hamham91:
@emma-floored @hwigmore Added another commit which passes `onKeyDown` to the text input (can you tell that I'm working with TextFields?)

emma-floored:
👍

hwigmore:
👍

PR #22, opened 2018-07-19
Federicoreyesgomez/ch9076/move-blocks-variables-out-of-stylus-make

frg100:
@emma-floored @jforzano This seems to be a working version from what I can tell, all the variables seem to be the same. I'm not exactly sure how to test it so maybe you guys could give me some suggestions there, but it looks functional. All the processing happens in gulp with 3 new npm packages that I installed


gulpfile.js
```diff
@@ -72,6 +76,28 @@ gulp.task('dev:react', () => {
...
+          .pipe(concat('variables.styl'))
+          .pipe(gulp.dest('./blocks-styles'));
+});
+
+gulp.task('dev:jsonToStylus', ['stylusGeneration'], () => {
+  return gulp.src(['./blocks-styles/variables-base-fonts.styl', './blocks-styles/variables.styl', './blocks-styles/variables-base-root.styl'])
+          .pipe(concat('variables.styl'))
+          .pipe(gulp.dest('./blocks-styles'));
+
+})
```
jforzano: What exactly is this part doing? I'm having trouble following

package-lock.json
```diff
@@ -7583,9 +7772,9 @@
       "integrity": "sha1-9+572FfdfB0tOMDnTvvWgdFDH9E=",
       "dev": true,
       "requires": {
-        "camelcase": "^1.0.2",
-        "cliui": "^2.1.0",
-        "decamelize": "^1.0.0",
+        "camelcase": "1.2.1",
+        "cliui": "2.1.0",
+        "decamelize": "1.2.0",
```
jforzano: @emma-floored are all these updates bad? It seems like a lot...

emma-floored: I guess some questions could be: do we need these packages? what exactly are the `gulp-json-sass` and `gulp-json-stylus` packages doing, and is it sufficiently annoying to do ourselves that we definitely want the package? it could be.

gulpfile.js
frg100: @emma-floored @jforzano Basically what these packages do is that they integrate with gulp and take in a json file, spitting our either sass or stylus. dev:jsonToSass is just calling the package and writing the results to ./blocks-styles. For the stylus, however, there were 2 parts that couldnt be automatically translated: some font stuff and ":root {}". These I separated into different stylus files. To include them both in the final stylus file, I generate a stylus file with the rest of the variables (stylusGeneration) and then add the fonts first, then the stylus that was derived from json, and then the ":root{}" part. Concat takes multiple files and spits out one file with all of them concatenated.

frg100: I guess in theory these packages aren't so complicated; all they do is go through the json file and format each variable according to the destination requirements, but I chose to use them because they were easy to use and quick to implement in a few lines

emma-floored:
👍 everything still works and the generated files look reasonable

jforzano:
looks good to me! after we merge this we have a ticket to pull in blocks changes and use the blocks variables sass sheet and get rid of our copied one. @frg100 can you ticket that?

jforzano:
👍

PR #23, opened 2018-07-19
remove findDOMNode

emma-floored:
@hwigmore @hamham91 this started breaking everything for some reason


react/lists/SortableList.jsx
```diff
@@ -36,10 +36,11 @@ class SortableList extends React.Component {
   }
 
   componentDidMount() {
-    this.listDOM = ReactDOM.findDOMNode(this);
-    this.listDOM.addEventListener('dragstart', this.onDragStart);
-    this.listDOM.addEventListener('dragover', this.onDragOver);
-    this.listDOM.addEventListener('dragend', this.onDragEnd);
+    if (this.listDOM) {
```
hwigmore: would this ever not exist and if so when / how would these event listeners get added?

hwigmore:
i had a question but overall makes sense and 👍

react/lists/SortableList.jsx
emma-floored: I don't think it should ever not exist but I also didn't think the bug we were having in datavis should ever have happened... so I figured maybe at least in this case it wouldn't crash even if the list doesn't work

hwigmore: ok!

PR #24, opened 2018-07-19
implemented dropdown text field

Jinsoo94:
Implemented dropdown text field. It inherits TextField and Dropdown.

https://app.zeplin.io/project/5adf8fe14c3924b93d974cba/screen/5b438b2c2f5f03c05f29cb54


react/input/DropdownTextField.jsx
```diff
@@ -0,0 +1,84 @@
+const React = require('react');
+const PropTypes = require('prop-types');
+const TextField = require('./TextField.jsx');
+const Dropdown = require('../dropdowns/Dropdown.jsx');
+
+class DropdownTextField extends React.Component {
+  constructor(props) {
+    super(props);
+
+    console.log('DropdownTextField created...');
```
emma-floored: can remove? actually, looks like you can remove the whole constructor

hwigmore:
this is really picky, but for clarity's sake i'd vote to rename this component something like `TextFieldWithDropdown` or `TextFieldDropdownCombo`. i think (though it could be just me) that makes it more obvious what the component is.

react/input/TextFieldWithDropdown.jsx
```diff
@@ -0,0 +1,78 @@
...
+          value={this.props.valueTF}
+          isValid={this.props.isValid}
+          invalidErrorMessage={this.props.invalidErrorMessage}
+          placeHolder={this.props.placeholder}
+          type={this.props.type}
+          name={this.props.name}
+          onChange={this.props.onChangeTF}
+        />
+        <Dropdown
+          text={this.props.text}
```
hwigmore: one thing that might be nice and make it more clear which of the props go to where (and to avoid having to use DD and TF) would be to have the props types be like 
```
textField: PropTypes.shape({
  label: ...,
  value: ...,
  onChange: ...,
  etc.
}),
dropdown: PropTypes.shape({
  options: ...,
  value: ...,
  etc
})
```

and then here where you use them you could simply do 

```
<TextField {...props.textField} />
<Dropdown {...props.dropdown />
```

react/input/TextFieldWithDropdown.jsx
```diff
@@ -0,0 +1,78 @@
+const React = require('react');
+const PropTypes = require('prop-types');
+const TextField = require('./TextField.jsx');
+const Dropdown = require('../dropdowns/Dropdown.jsx');
+
+class DropdownTextField extends React.Component {
+  render() {
```
hwigmore: if a react component only has a `render` function and nothing else, you can write it as a pure function, instead of an extension of React.Component! so like 

```
function DropdownTextField(props) {
  return (
    .. all the things ..
  );
}

react/input/TextFieldWithDropdown.jsx
hwigmore: (not on this pr but eventually we might want the concept of shared prop types. we do this in spacer. and then we wouldn't have to duplicate all of these text field props and dropdown props)

hwigmore:
looks like your changes to preview.js didnt get committed. if you run gulp server again you'll see preview.js gets changed. add and commit that change and then this is good to go from me 👍 nice work!!!

PR #25, opened 2018-08-03
update font to graphik, separate font definitions out to their own fo…

emma-floored:
…lder

Updating the font to Graphik now that we have paid for it!

@MagdalenaHenke @jforzano @hwigmore @joshkarges maybe?

A lot of these changes are tweaks for the blocks documentation, but one thing I did is put the font specification css file into the fonts/ folder along with the font files. It gets pulled into the final blocks.css file anyway, so I don't think it should make a difference? But that's a change I made.

Biggest thing to pay attention to as far as conflicts might go is that I got rid of some of the font family variables (originally there was like, `$sans-serif-font-family-normal`, `$sans-serif-font-family-bold`, etc, now there's only `sans-serif-font-family` which is `'Graphik Web'` and you control the weight via things like `font-weight: $normal_weight` or `font-weight: $bold_weight` (or `style: italic` if you're using any italics).

But look now the text is centered in our buttons, yay!:
screen shot 2018-08-03 at 2 38 56 pm


emma-floored:
https://app.clubhouse.io/floored/story/9526/update-fonts

hwigmore:
woo! 👍

PR #26, opened 2018-08-03
Prefix with blx

emma-floored:
@MagdalenaHenke @jforzano @hwigmore @joshkarges @vladcbre 

This one is a doozy. If you were using any Blocks class names outside of Blocks, beware, because they have changed (this will be rough for Datavis 😭).

But, if you were not using Blocks, like Plans wasn't, because some of the class names conflicted with your class names, you should have no problems and now the classes should all be unique.

I tested all the pages and tried to be careful but it is still possible I missed something 😬


emma-floored:
https://app.clubhouse.io/floored/story/9077/prefix-blocks-classes-with-blx

hwigmore:
did a quick scan and i noticed you didnt prefix things like `.disabled` and `.hidden` which seem especially prone to overlap with other app-specific classes. should we prefix everything in blocks?

emma-floored:
@hwigmore good point. I think originally I was thinking those classes might not be too different so maybe it was ok, but actually now that you mention it, it's probably better for just everything to be prefixed with `blx`

hwigmore:
yeah i had the same thought for a sec like is this _really_ necessary but there are some specific opacity things, for instance, on disabled so might as well keep it separate!

emma-floored:
@hwigmore I have a few classes that are mostly used to "fake" events for showing all the styles on the doc pages (`.hover` and `.focus` for `:hover` & `:focus`) - do you think those should be prefixed as well or left as is? They are never used at a root level, only nested in other `blx-` components.

hwigmore:
oh interesting. my instinct is to just `blx` it all for consistency

emma-floored:
updated!

hwigmore:
👏 👏 ! 

i think it'll be more efficient for us to update any potentially missed things after merge, rather than reading through the changes on this, so i'm gonna say 👍

jforzano:
agree with all the things said!

jforzano:
👍

emma-floored:
I'm gonna go ahead and say 2 👍 is sufficient :P

PR #29, opened 2018-08-07
add LinkButton and dry up preview.jsx

hwigmore:
@emma-floored @MagdalenaHenke 

we need a link that looks like a button in Spacer, and i suspect this need will come up more in the future / for other apps. 

i added `LinkButton.jsx` which the various button components use if passed an `href` rather than an `onClick`

i then went to add examples to the react preview page, and while i was in `preview.jsx`, i realized i could dry up a lot by making a `getPreviewComponent` function (also feels a little easier to read through, i think) so i went ahead and did that, too

note: i think the link button and the button button look 99% alike but there seems to still be a little difference in font weight/spacing? that must be from built in browser styles. decided to leave it for now since it's so minor and wasn't easy to track down


hwigmore:
oh also @runigoswami just wanted to shout out and let you know i've implemented this for now. 

i'm not sure if we want to have a larger discussion around whether or not we _should_ have links that look like buttons, but since it's a common pattern in our apps right now, i went ahead and added them (last row)

screen shot 2018-08-07 at 1 56 26 pm

emma-floored:
I was actually thinking yesterday about how this would be nice for Datavis... 💐 

code looks good to me, I'll test it in a bit

docs/_javascript/preview.jsx
```diff
@@ -34,6 +34,20 @@ const SortableList = BlocksReact.SortableList;
 
 const INVALID_VALUE = 'Wrong';
 
+
+function getPreviewComponent(caption, component) {
```
emma-floored: ++

react/buttons/LinkButton.jsx
```diff
@@ -0,0 +1,30 @@
+const React = require('react');
+const PropTypes = require('prop-types');
+
+
+const LinkButton = props => (
+  <a
+    className={`blx-button ${props.classes} ${props.isDisabled ? 'blx-disabled' : ''}`}
+    href={props.href}
```
MagdalenaHenke: sooooo - I think it's worth thinking about actually using Ract Links (with `to` properties) for all of blocks.

MagdalenaHenke: We could make two different Link buttons - one for app internal routing, that's based on the React Link (and keeps the things that are on the store intact) and one for external routing, that's an actual `<a` with an href

hwigmore: interesting. that would only be for internal links but external would still need regular `<a>` right?

hwigmore:
@emma-floored @MagdalenaHenke 

@runigoswami says she doesn't love this idea, so we might need to hold on this and instead

1. make app-specific adjustments for link buttons, like we do now 
2. ask designers to redesign any link buttons so they dont look like buttons
3. something else?

react/buttons/LinkButton.jsx
hwigmore: oh sorry your second comment hadnt loaded when i wrote my response!

emma-floored:
2 would be ideal I think

MagdalenaHenke:
What made you chose to have the user specify `useReactLink` over `useHref`/`externalLink` or something? 

I'm wondering because it seems to me that the vast majority of links in the app will be internal links which by good practice should be React Links (now that I type that out, it makes me question whether it's true that most links are internal), and also I think it'll be easy to forget to use `useReactLink` since it'll still work (as in, it'll still route you to a different page) but then cause some bug around why the store got reset.

I can't think of a situation where we'd want an internal <a > - so we could also use the already existing `isExternal` prop to decide which one to use.

hwigmore:
I chose to make that explicit because I think we don't want to make any assumptions in Blocks about the context in which these components are being used. Specifically, we don't know if the app using Blocks has React Routing set up or not. So to me the most basic level is a generic `<a>` but any apps using React Router can choose to use the associated React Link. 

For something like Spacer, which does use React Router, we could make an app-specific component that wraps around this Blocks LinkButton and always passes in `useReactLink`

hwigmore:
@emma-floored @MagdalenaHenke merged this and pushed a follow up right to master (`preview.jsx` changes hadnt been checked in. i dont understand how that happened!)

PR #30, opened 2018-08-13
after close on click

MagdalenaHenke:
@hwigmore @emma-floored 

This is for a branch open in Spacer - we added our own custom dropdown trigger, where the trigger is a text field (it's for a location dropdown that shows location suggestions as you type). I wanna be able to clear the text field if the dropdown got abandoned (so if they type some junk in, "blabla", for example, and then click outside of the dropdown somewhere and close the dropdown that way). I thought that's probably general enough that it's ok to add to blocks itself.


hwigmore:
you might need to `rm docs/lib/preview.js` and rerun gulp locally to get the new changes since the branch you made this on wasn't up to date? i'm still working on the branch bringing workplace up to date but it has a lot of side tasks, sorry!!

MagdalenaHenke:
I rebased onto the current blocks master and updated preview.js. If/when this gets merged, I'll squash&merge cause I don't wanna deal with getting rid of the first commit on here right now :)

hwigmore:
tests good looks good 👍

PR #31, opened 2018-08-13
add modal to react preview and simplify jsx component

hwigmore:
@emma-floored with the `blx-` prefixes, the modal close button is no longer properly positioned on spacer. it was not in the upper right (i guess previously it worked because we had classname overlaps between blocks and bulma). 

this probably wasn't caught because modal wasn't in preview yet. i realize modals aren't speced out yet, so i didn't add/update the ActionModal one, but just got BaseModal looking ok by moving the close button, and i added it to preview.jsx

i also did a tiny bit of refactoring for clarity, because i found it easier to see `if (props.isHidden) return null;` and then not having to conditionally define other things based on hidden status

the PR for integrating blocks changes into spacer is already feeling big / hard to manage, so i just made this straight on blocks

@MagdalenaHenke @jforzano would you mind taking a look as well in case emma is too busy siggraphhing?


docs/_javascript/preview.jsx
```diff
@@ -31,6 +31,8 @@ const LinkTabs = BlocksReact.Tabs.LinkTabs;
 
 const SortableList = BlocksReact.SortableList;
 
+const BaseModal = BlocksReact.Modals.BaseModal;
+const ActionModal = BlocksReact.Modals.ActionModal;
```
hwigmore: oh i can remove this for now

hwigmore:
ah i will fix my conflicts and f-push. @MagdalenaHenke @jforzano will tag you when ready again

hwigmore:
@MagdalenaHenke @jforzano @emma-floored ready again!

jforzano:
looks good to me! 👍

PR #33, opened 2018-08-16
fixed modal instead of absolute

hwigmore:
@jforzano this will fix the issue on spacer pro with the modal background, so it's fixed to the viewport not the closest relative parent


jforzano:
cute cute cute cute cute

(to the tune of "work work work work work")

👍 !

PR #35, opened 2018-08-21
split react preview components out into separate files

emma-floored:
@hwigmore 

I think this could be considered part of the ticket you made for eventually including react previews on regular doc pages. I split out each chunk of the big `preview.jsx` file into its own file, which will also hopefully make updating that page less painful in the future.


docs/_javascript/AccordionPreview.jsx
```diff
@@ -0,0 +1,29 @@
+const React = require('react');
+
+const BlocksReact = require('../../react-index.js');
+const AccordionSection = BlocksReact.Accordion;
```
hwigmore: this is _really_ not a big deal at all but in all these files you might as well just 
`const AccordianSection = require('../../react-index').Accordian` so you aren't importing way more than is necessary

emma-floored: oh yeah, I was just lazy and copied from when it was all in one file, but your way is better :P

emma-floored:
@hwigmore made both those changes!

hwigmore:
👍

hwigmore:
thank you for doing this, by the way!!

PR #36, opened 2018-08-23
have button component use children instead of text attribute so it's …

emma-floored:
@hwigmore @MagdalenaHenke @jforzano want to be sure none of you like, hate this change

Recently, @Jinsoo94 needed a button with both text and an icon inside. Since the current button component only accepts `text` as an attribute, he had to make his own. I thought it'd be nice for this button component to be more flexible, so I got rid of `text` and instead have the button use `children` for its contents.


hwigmore:
that works! my only concern is that blocks buttons will have a lot more potential for variability depending on the the children passed in. i wonder if we should explicitly pass a `rightIcon` or `leftIcon` and do the icon sizing/spacing in blocks. 

but also fine with this for now and we can come back to that suggestion.

emma-floored:
Oh interesting... I think I like that idea better.

emma-floored:
Going to close this and re-open after I implement Heather's suggestion

PR #37, opened 2018-08-23
Tweak button component

emma-floored:
@hwigmore buttons with icons, updating a few styles to match the current button spec, button groups (wrap buttons in a div with class `blx-v-button-group` or `blx-h-button-group`)


emma-floored:
screen shot 2018-08-23 at 4 50 51 pm

hwigmore:
nice! 👍 

one thought i had is that it's possible one day we'll want to let apps include their own icons, not only ones that are already checked into blocks, but for now this is great!

PR #38, opened 2018-08-24
arrays of components need keys, too

hwigmore:
@emma-floored tiny tiny but i didnt realize arrays need keys, too


emma-floored:
👍

PR #39, opened 2018-08-24
React on docs

emma-floored:
@hwigmore @jforzano @MagdalenaHenke @Jinsoo94 

This is maybe not perfect just yet but it does work and I wanted to get some eyes on it. A lot of files look like they changed here, but I'd say the most important one to look at is the gulpfile. Basically, instead of having 1 jsx file (`preview.jsx`) that included all the react components, I now bundle each component preview into its own individual script (plus a separate vendor file for React), and then include that script on the specific doc page for whatever component (so, `docs/_javascript/ButtonPreview.jsx` -> `docs/lib/ButtonPreview.js` which is included in `docs/components/buttons.html`).

screen shot 2018-08-24 at 5 23 32 pm

There is one thing I tried out but didn't include in this PR yet - source maps. They seem helpful for debugging, but I wasn't sure how to not include them on the production Blocks site.


emma-floored:
https://app.clubhouse.io/floored/story/10002/refactor-react-previews-so-each-component-s-react-preview-lives-alongside-its-main-component-definition

hwigmore:
yess thank you for this! one thing is i'm not sure watchify is working because i tried changing a react preview file and didnt see the change even on refresh. but still, 👍

hwigmore:
actually, i just noticed something that would be good to avoid if possible. when i make an update to one of the `Preview.jsx` files, after re-running all the gulp tasks, every `Preview.js` files is updated

i added a console log to `DropdownPreview.jsx` and now `git status` shows

screen shot 2018-08-25 at 10 28 06 am

gulpfile.js
```diff
@@ -78,10 +84,70 @@ gulp.task('dev:jekyll', () => {
...
+  './docs/_javascript/CheckboxPreview.jsx',
+  './docs/_javascript/DropdownPreview.jsx',
+  './docs/_javascript/ModalPreview.jsx',
+  './docs/_javascript/ProgressPreview.jsx',
+  './docs/_javascript/RadioButtonPreview.jsx',
+  './docs/_javascript/SortableListPreview.jsx',
+  './docs/_javascript/TabPreview.jsx',
+  './docs/_javascript/TextFieldPreview.jsx',
+  './docs/_javascript/TogglePreview.jsx'
+];
```
jforzano: Would be cool to have this auto generated somehow, or if not, maybe moved somewhere external that makes it more apparent that when you add a preview it needs to be listed here or something? not like super necessary but just a thought!

emma-floored: agree... I'll see if I can figure something out.

emma-floored:
@hwigmore @jforzano I tried to make the gulp tasks a little better. I think it is definitely still doing a little bit of extra work, but it no longer regenerates all the component bundles when you modify a component (although it may regenerate multiple - for example, if you change the primary button, that affects the modal bundle as well as the button bundle).

I also installed a package that let me get rid of the file list, so we wouldn't have to keep that up to date anymore.

hwigmore:
yay!! not blocking but out of curiosity, if you make a change is it going through for you without restarting gulp? it's not for me. but def 👍 this is huge improvement having these react components more accessible!!

jforzano:
great stuff! 👍

emma-floored:
@hwigmore I did see updates... can you tell me which file you tried editing that didn't update? Maybe I missed a directory or something.

emma-floored:
ok, I'll check it out

emma-floored:
@hwigmore Mine does update. But, as I was fiddling, I realized I might not have mentioned that testing requires an `npm install` - did you do that & if not, does that fix the issue?

output will look like this:
screen shot 2018-08-31 at 12 17 40 pm

hwigmore:
so confused why mine isn't working! i `npm install`ed and when i change something in the preview file i get the first part "xxx bytes written (yyy seconds)" but not the regenerating part! 

but again i dont think this should hold up merge so go for it and i can look into this separately

PR #41, opened 2018-09-06
change black text on black color card

hwigmore:
Clubhouse: https://app.clubhouse.io/floored/story/9275/documentation-can-t-read-text-01-in-the-color-block-because-the-text-color-matches-the-background-color


hamham91:
💯

PR #43, opened 2018-09-10
Component updates

emma-floored:
@hwigmore @MagdalenaHenke @jforzano @Jinsoo94 

Updates to several different blocks components based on Runi's list of stuff to do for v1!

I updated:
- check boxes
- radio buttons
- tabs
- links
- buttons
- dropdowns
- alerts
- text fields

(but I did not do all of the updates for each of these components - just the easiest ones for now)


emma-floored:
[ch10098]
[ch10103]
[ch10105]
[ch10100]
[ch10097]

Jinsoo94:
wow these look super nice! I especially like the new radio styles and alert styles.

blocks-styles/buttons.styl
```diff
@@ -78,12 +78,23 @@
...
     text-decoration: none;
-    box-shadow: 0 2px 4px 0 $button-hover-shadow-color;
+    box-shadow: 0 1px 3px 0 $button-hover-shadow-color;
 
     &.blx-primary {
       border: 2px solid $primary-button-focus-border-color;
       background-color: $primary-button-focus-background-color;
       color: $primary-button-focus-text-color;
+      
+      &::before {
```
hwigmore: what is this guy doing?

docs/_data/navigation.yml
```diff
@@ -74,6 +74,5 @@
   subsection: Navigation
 
 - title: Tabs
-  url: components/tabs
-  section: Components
-  subsection: Navigation
+  url: patterns/tabs
```
hwigmore: i don't understand what makes this a "pattern". do you?

blocks-styles/buttons.styl
emma-floored: If you look at the spec, the primary button focus state has like a double outline and also a drop shadow (so does the checkbox, for that matter). I believe you can't actually do a double border, unless you fake it with an inset shadow, but we're already doing a shadow, so I faked it instead with a pseudo-element.

docs/_data/navigation.yml
emma-floored: no

emma-floored:
@hwigmore re: your lib files changing, the only React component that changed in this PR was TextField, so the diff looks correct for now (only `docs/lib/TextFieldPreview.js` checked in). Not sure what happened with yours...

hwigmore:
ughhh this is the same thing as before i guess. i really hoped that would magically go away. will look into it

blocks-styles/buttons.styl
hwigmore: clever!

hwigmore:
ok sorry for slacking on this. now that i understand why i see those diffs, going back to actually review

Jinsoo94:
@hwigmore I also noticed that the files in docs/lib/ got altered when I ran gulp server. Let me know if you figure out how to address it!

hwigmore:
@Jinsoo94 i figured it out! once this is updated with master that problem should go away!

hwigmore:
this was the change if you are curious https://github.com/floored/blocks/pull/45 basically we were getting our personal computer paths like (heather/documents/...) when we bundled!

hwigmore:
ok i'm sorry @emma-floored i'm going to give this a 👍 because it looks good at first glance and reviewing css is hard and we can tweak if we find necessary things as we go!

emma-floored:
no worries! thanks for looking 💐

PR #44, opened 2018-09-10
popover menus

emma-floored:
@hwigmore @MagdalenaHenke @Jinsoo94 

New Block - popover menus

screen shot 2018-09-10 at 6 46 07 pm


emma-floored:
[ch10102]

hwigmore:
oooh i've been making one of these myself in spacer pro. i wonder if runi would be open to a variation that has the three dots vertical 😈

emma-floored:
@hwigmore maybe! if so, you should be able to do it using this one by just defining a new icon class that is the 3 vertical dots.

hwigmore:
ok runi said she's chill with us making it so that any icon can be passed in and used as the popover icon, so i might add that once this is in. reviewing now

react/dropdowns/PopoverMenu.jsx
```diff
@@ -0,0 +1,77 @@
...
+        onClick={props.toggle}
+      >
+        <span className={`blx-icon blx-icon-${props.icon}`} />
+      </button>
+      <ul className={`blx-dropdown-menu ${props.isOpen ? '' : 'blx-hidden'} ${props.isLeft ? 'blx-popover-is-left' : 'blx-popover-is-right'}`}>
+        {
+          props.options.map((option) => {
+            let item = null;
+            if (option.href) {
+              item = (
```
hwigmore: not a necessary change here but just an FYI, on the branch where i working on a similar thing, i included an `option.element` for when something special needs to be passed in https://github.com/floored/workplace/blob/user-action-menu/ch10037/assets/js/spacerPro/components/_common/form/CustomTriggerDropdown.jsx#L9-L37 and i wonder if that would be allowed as a blocks level thing

(in our case it was a custom react element that copies to clipboard https://github.com/floored/workplace/blob/user-action-menu/ch10037/assets/js/spacerPro/components/users/UserInviteTagAndActionMenu.jsx#L23-L37)

react/dropdowns/PopoverMenu.jsx
```diff
@@ -0,0 +1,77 @@
...
+    </div>
+  </div>
+);
+
+PopoverMenu.propTypes = {
+  text: PropTypes.string,
+  isOpen: PropTypes.bool,
+  isLeft: PropTypes.bool,
+  options: PropTypes.arrayOf(PropTypes.shape({
+    text: PropTypes.string,
```
hwigmore: required?

hwigmore:
👍

react/dropdowns/PopoverMenu.jsx
emma-floored: that sounds good... I'll add it to DropdownMenu as well b/c I just reached a point in Datavis where I want that... (I'm doing a very gross thing where I am trying to resolve all the Datavis conflicts with Blocks but with the most updated stuff on this branch by copying this branch's files into the node_modules of Datavis... it's been good for testing these 2 PRs, I guess!)

react/dropdowns/PopoverMenu.jsx
emma-floored: No. Some of our popover menus are just icons.

hwigmore: this is for the option text though. although i guess based on my suggestion above of `option.element`, `option.text` might not be required (unless necessary for a key)?

emma-floored: ohhh duh...

react/dropdowns/PopoverMenu.jsx
hwigmore: how would it work as children if some are special elements but some are defined in the `options` array?

emma-floored: omg my brain is just dead

emma-floored: I was thinking special element for like the dropdown trigger part

emma-floored: rather than as an option in the menu

hwigmore: ooooh yeah i was talking about option in the menu!

PR #45, opened 2018-09-12
remove full paths so we dont get each devs path when they re-bundle

hwigmore:
@emma-floored @MagdalenaHenke would you run `gulp server` locally from this branch and confirm no files change? 

also, @emma-floored was there a reason you had that opt set to `true`?


emma-floored:
👍 I don't remember... I think I must have thought it was for something else

PR #46, opened 2018-09-13
keyboard interaction with the following components

Jinsoo94:
Copied over changes made in datavis. There will not be a visible change on gulp server because to have a keyboard interaction, event handler must be passed down via onKeyUp or onKeyDown. However, dropdown interaction is defined within the component, so its new behavior will be visible on gulp server. @emma-floored @hwigmore  

1. TextField
2. Dropdown (arrow key, tab key)
3. Checkbox
4. Button


hwigmore:
i'm a little confused by your distinction of what changes are visible vs not visible on gulp server. what do you mean by that?

Jinsoo94:
For TextField.jsx, Checkbox.jsx, and Button.jsx, all I did is adding new props so it will not change anything you see on localhost:8080. However, Dropdown.jsx has a new event handler function written in the file, and its effect is visible on localhost:8080.

hwigmore:
ah got it, thanks!

react/dropdowns/Dropdown.jsx
```diff
@@ -6,11 +6,49 @@ const closeOnClick = require('../wrappers/closeOnClick.jsx');
...
+    
+    this.state = {
+      onFocusIdx: -1
+    };
+    
+    this.optionsNodes = this.props.options.map(() => null);
+    
+    this.setOptionNode = (element, idx) => {
+      this.optionsNodes[idx] = element;
+    }
```
hwigmore: i think the react standard is to only define

`this.fn = `

up in the `constructor` if you are needing to `.bind(this)` for scope reasons (like line 22). otherwise, component functions should be defined outside of the constructor, like how `onSelect` is done

react/input/TextField.jsx
```diff
@@ -61,7 +61,7 @@ class TextField extends React.Component {
         onKeyUp={(e) => {
           if (this.props.onKeyUp) this.props.onKeyUp(e);
           if (e.keyCode === 13) { // ENTER
-            e.target.blur();
+            e.target.blur(); // disable blur if you want to keep the focus on this input
```
hwigmore: since this PR introduces `e.key` instead of `e.keyCode`, can we change line 63 to be consistent?

also, re: this new comment. is that something we'll want the option to do? if so, we should have a prop on this component that's like `shouldBlurOnEnter` and it defaults to true but can be false. otherwise, i'm not sure what the comment really means, action wise

react/dropdowns/Dropdown.jsx
```diff
@@ -6,11 +6,49 @@ const closeOnClick = require('../wrappers/closeOnClick.jsx');
...
   onSelect(option) {
     if (this.props.onChange) this.props.onChange(option);
     this.props.close();
   }
 
+  arrowKeyHandler(e) {
+    if (e.key === 'ArrowDown' || e.key === 'ArrowUp') e.preventDefault();
+    if (!this.props.isOpen && e.key === 'ArrowDown') this.props.open();
+    if (this.props.isOpen && e.key === 'ArrowDown') {
+      this.setState({ onFocusIdx: Math.min(this.state.onFocusIdx + 1, this.props.options.length - 1) }, () => {
```
hwigmore: something that we didn't do originally but have started trying to do more, after @MagdalenaHenke realized it's a React best practice, is updating state that depends on previous state like this https://teamtreehouse.com/community/react-docs-now-recommends-using-function-with-prevstate-inside-of-setstate

react/buttons/Button.jsx
```diff
@@ -7,6 +7,10 @@ const BasicButton = props => (
     className={`blx-button ${props.classes} ${props.isDisabled ? 'blx-disabled' : ''}`}
     onClick={props.onClick}
     disabled={props.isDisabled}
+    autoFocus={props.autoFocus}
+    onKeyDown={props.onKeyDown}
+    onKeyUp={props.onKeyUp}
+    ref={props.setRef}
```
hwigmore: out of curiosity, what are the use cases for this? i think i've only ever defined a ref in the context of a component, but never passed in a function for setting it from the outside

react/dropdowns/Dropdown.jsx
```diff
@@ -6,11 +6,49 @@ const closeOnClick = require('../wrappers/closeOnClick.jsx');
...
+  }
+
   onSelect(option) {
     if (this.props.onChange) this.props.onChange(option);
     this.props.close();
   }
 
+  arrowKeyHandler(e) {
+    if (e.key === 'ArrowDown' || e.key === 'ArrowUp') e.preventDefault();
+    if (!this.props.isOpen && e.key === 'ArrowDown') this.props.open();
```
hwigmore: this might be a personal preference, or a question for product, but i would want the first down key to highlight the first option. right now you have to hit down twice to accomplish that. could combine lines 32 and 33 with 

```
if (e.key === 'ArrowDown') {
  if (!this.props.isOpen) this.props.open();
  ...all the rest of the stuff starting on line 34...
}
```

react/dropdowns/Dropdown.jsx
```diff
@@ -6,11 +6,49 @@ const closeOnClick = require('../wrappers/closeOnClick.jsx');
...
+    if (this.props.isOpen && e.key === 'ArrowDown') {
+      this.setState({ onFocusIdx: Math.min(this.state.onFocusIdx + 1, this.props.options.length - 1) }, () => {
+        this.optionsNodes[this.state.onFocusIdx].focus();
+      });
+    }
+    if (this.props.isOpen && e.key === 'ArrowUp') {
+      this.setState({ onFocusIdx: Math.max(this.state.onFocusIdx - 1, 0) }, () => {
```
hwigmore: something that i think would help with readability would be to define these actions as component functions using clear language so you could have something like (psuedo code)

```
if (key == ArrowDown) {
  if (notOpen) open();
  this.focusOnNextOption();
}

if (key === ArrowUp) {
  this.focusOnPreviousOption();
}

**Comment by Jinsoo94** [comment]
@hwigmore hi sorry I was caught up with datavis work last week so couldn't go over last week. I will go over these today!

**Comment by hwigmore** [comment]
no rush! just wanted to make sure i wasn't confusing you!

File: react/buttons/Button.jsx (hunk 4 above)
Comment (Jinsoo94) [reply react/buttons/Button.jsx]: In datavis modals, we want the 'Next' Button to be focused when the user presses enter key while on the textfield. That is why we need ref in here!

Example:
SomeFileThatUsesButton.jsx

constructor(props) {
  super(props);
  this.nextButtonNode = null;

  this.setNextButtonRef = this.setNextButtonRef.bind(this);
  ...
}

setNextButtonRef(element) {
  this.nextButtonNode = element;
}

textFieldEnterKeyHandler() {
  this.nextButtonNode.focus();
}

...
render() {
  return(
    ...
    <Button
      ...
      setRef={this.setNextButtonRef}
    />
  );
}

Comment (emma-floored) [reply react/buttons/Button.jsx]: So I was reading about this, and I think we could get rid of this `ref` here if we make `Button` a class instead of a pure component. That _should_ let us do like `<Button ref={blah} />` from the parent instead of having it this way.

Comment (Jinsoo94) [reply react/buttons/Button.jsx]: ohhhhh right. I have definitely ran into this issue before but forgot that you cannot set ref to a functional component.

**Comment by hwigmore** [comment]
sorry @Jinsoo94 not trying to ignore you, will review your updates later today or tomorrow!

File: react/input/TextField.jsx [hunk 7]
```diff
@@ -110,7 +111,9 @@ TextField.propTypes = {
   onFocus: PropTypes.func,
   onBlur: PropTypes.func,
   autoComplete: PropTypes.string,
-  readOnly: PropTypes.bool
+  readOnly: PropTypes.bool,
+  autoFocus: PropTypes.bool,
+  setRef: PropTypes.func
```
emma-floored: so theoretically, since TextField is already a class, we shouldn't need this, right? should be able to set a ref on the TextField from within it's parent?

Jinsoo94: I would guess so? I will try it and let you know.

react/dropdowns/Dropdown.jsx
```diff
@@ -6,11 +6,57 @@ const closeOnClick = require('../wrappers/closeOnClick.jsx');
...
 
 class Dropdown extends React.Component {
+  constructor(props) {
+    super(props);
+    
+    this.state = {
+      onFocusIdx: -1
+    };
+    
+    this.optionsNodes = this.props.options.map(() => null);
```
hwigmore: can we just start this as `this.optionsNodes = []`? not sure if it's important that it starts off equaling the length of the actual options, since it will get updated / set as those option elements are built

react/dropdowns/Dropdown.jsx
```diff
@@ -6,11 +6,57 @@ const closeOnClick = require('../wrappers/closeOnClick.jsx');
...
+    
+    this.arrowKeyHandler = this.arrowKeyHandler.bind(this);
```
hwigmore: this isn't specific to arrow keys (it gets called on all key downs, and has specific instructions for the tab key as well). can we just call it `this.onKeyDown`?

react/dropdowns/Dropdown.jsx
```diff
@@ -59,6 +109,12 @@ class Dropdown extends React.Component {
                   item = (
                     <button
                       onClick={() => this.onSelect(option)}
+                      onKeyDown={e => {
+                        if (e.key === 'Enter') e.preventDefault();
```
hwigmore: is there a reason to have this `e.preventDefault()` here, instead of including `Enter` on line 35?

react/dropdowns/Dropdown.jsx
```diff
@@ -6,11 +6,57 @@ const closeOnClick = require('../wrappers/closeOnClick.jsx');
...
     if (this.props.onChange) this.props.onChange(option);
     this.props.close();
   }
 
+  setOptionNode(element, idx) {
+    this.optionsNodes[idx] = element;
+  }
+
+  arrowKeyHandler(e) {
+    if (e.key === 'ArrowDown' || e.key === 'ArrowUp' || e.key === 'Tab') e.preventDefault();
```
hwigmore: not necessary but if you wanted to shorten this and make it easier to add on in the future you could do 
`if (['ArrowDown', 'ArrowUp', 'Tab', 'Enter'].indexOf(e.key) > -1) e.preventDefault()`

react/dropdowns/Dropdown.jsx
```diff
@@ -6,11 +6,57 @@ const closeOnClick = require('../wrappers/closeOnClick.jsx');
...
+    }, () => this.focusOption(this.state.onFocusIdx))
+  }
+
+  getFocusIndexAfterKeyPress(keyName) {
+    if (keyName === 'ArrowDown') return Math.min(this.state.onFocusIdx + 1, this.props.options.length - 1);
+    if (keyName === 'ArrowUp') return Math.max(this.state.onFocusIdx - 1, 0);
+    if (keyName === 'Tab') {
+      const focusIsAtTheEndAndShouldWrapToBeginning = this.state.onFocusIdx === this.props.options.length - 1;
+      if (focusIsAtTheEndAndShouldWrapToBeginning) return 0;
+      else return this.state.onFocusIdx + 1;
```
hwigmore: no need for this `else`! can just return

hwigmore:
actually two other things: 
1. there are conflicts to resolve 
2. the changes to the preview files didn't get checked in (if you `gulp server` then `git status` you'll see uncommitted changes)

hwigmore:
i tested on the react preview of dropdown, arrows up and down worked beautifully but then when i click enter on an option the entire dropdown disappears and i get errors in the console. let me know if you have trouble reproducing!

Jinsoo94:
@hwigmore ohh I didn't know I had to commit changes to the preview files. I will push it next time I commit. 

Dropdown error I think I know what you are talking about. I will look into it and let you know.

emma-floored:
oops that didn't fix everything, fixing again..

react/dropdowns/Dropdown.jsx
emma-floored: ooh so cute

emma-floored:
ok! I think I fixed it now

PR #47, opened 2018-09-14
dropdown overflow with ellipsis

hwigmore:
inspired by this in Spacer Pro. i was gonna make a local override but figured it'd be useful elsewhere. 

screen shot 2018-09-14 at 12 14 08 pm

added an example to the docs. includes `title` attribute on hover
screen shot 2018-09-14 at 2 38 50 pm

@emma-floored @MagdalenaHenke would you mind taking a look?


emma-floored:
oops! I actually thought we already had this

blocks.css
```diff
@@ -909,6 +909,8 @@ a.blx-button {
   flex-direction: column;
   min-width: 150px;
   max-width: 300px;
+  overflow: hidden;
+  text-overflow: ellipsis;
```
emma-floored: generally when I do this, I also add `white-space: nowrap;` but I dunno if it's actually necessary

hwigmore: oh yeah i usually do too but for some reason forgot it here!

hwigmore:
ok to merge?

emma-floored:
👍

PR #48, opened 2018-09-14
prefix suffix can be node

hwigmore:
for instance, when entering a password in Spacer Pro, after the min number of characters we show a checkmark. 

added a react preview example
screen shot 2018-09-14 at 3 12 38 pm

@emma-floored @Jinsoo94


react/input/TextField.jsx
```diff
@@ -96,10 +96,13 @@ TextField.propTypes = {
...
   label: PropTypes.string,
   placeholder: PropTypes.string,
-  prefix: PropTypes.string,
+  prefix: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.node
+  ]),
   suffix: PropTypes.oneOfType([
     PropTypes.string,
-    PropTypes.object
```
hwigmore: do you know if `PropTypes.object` would ever be needed / work how we have this set up? i think not?

emma-floored: oh I think I didn't know `.node` was a type and I used `.object` to also mean that

hwigmore:
ok to merge?

emma-floored:
👍

PR #49, opened 2018-09-17
remove tip about font relative pathing to blocks.css

hwigmore:
@emma-floored tiny thing. this tip i had added is no longer relevant. i thought about updating it to specify the pathing needed now, but decided it's a dumb thing to have to maintain, because we can just look at `blocks.css` and figure it out


emma-floored:
👍

PR #50, opened 2018-09-17
add in missing hover style for popover menus

emma-floored:
@hwigmore realized I missed hover styles on popover menus


hwigmore:
👍

PR #51, opened 2018-09-17
dropdown items need keys, not all have text

hwigmore:
@emma-floored to get rid of the react warning re: keys.

in the blocks preview examples, all dropdown options have a `value` so i could've made it that property. but we have examples in spacer where there's no value (for example, links) so i figured in those cases it was better to pass in a specific key. thoughts?


hwigmore:
@joshkarges

hwigmore:
@joshkarges @emma-floored i undid the overflow stuff. @emma-floored remember how you were like, i thought i already added that? you had. it was on `blx-dropdown-trigger`, where it should be. spacer must've just been on an outdated blocks so i didnt realize 😳

emma-floored:
oops!

hwigmore:
gonna merge this to fix dropdowns (and, less importantly, get rid of the console error). 🙏 i'm not ruining something else

PR #52, opened 2018-09-17
default to no placeholder

hwigmore:
@emma-floored are you ok with this? we've been passing in `placeholder=""` in spacer pro for no placeholder, but it seems more intuitive to me that you'd only pass if you want something displayed


emma-floored:
This makes sense to me 👍

PR #55, opened 2018-09-19
Input value is stored on the state for TextField component

joshkarges:
If the input tag has a `value` property, we need to update that property to reflect what the user has typed.

@emma-floored @hwigmore Am I missing something obvious here?

### DIFF
--- a/docs/_javascript/TextFieldPreview.jsx
+++ b/docs/_javascript/TextFieldPreview.jsx
@@ -18,12 +18,18 @@ class TextFieldPreview extends React.Component {
     super();
 
     this.state = {
-      conditionalSuffixValue: '',
       unitValue: 'ft',
-      date: null,
-      multiline: "On the eighty-fifth day of his unlucky streak, Santiago takes his skiff into the Gulf Stream, sets his lines and, by noon, has his bait taken by a big fish that he is sure is a marlin.",
+      defaultExampleValue: '',
+      defaultWithTextLabelValue: '',
+      filledExampleValue: 'Text input',
+      prefixSuffixExampleValue: '',
+      conditionalSuffixValue: '',
+      textFieldDropdownValue: '',
       validityExampleValue: INITIAL_INVALID_VALUE,
-      validityExampleIsValid: isFieldValid(INITIAL_INVALID_VALUE)
+      validityExampleIsValid: isFieldValid(INITIAL_INVALID_VALUE),
+      dateExampleValue: '',
+      iconExampleValue: 'Text input'
+      multilineExampleValue: "On the eighty-fifth day of his unlucky streak, Santiago takes his skiff into the Gulf Stream, sets his lines and, by noon, has his bait taken by a big fish that he is sure is a marlin.",
     };
 
     this.onChangeUnit = this.onChangeUnit.bind(this);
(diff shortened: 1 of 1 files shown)


hwigmore:
so i think there's no reason to update the react component, just the instance we use in the preview. i believe it's best practice to just display the props that are passed and let the parent/caller deal with the current state

for example, this is how we use the block's textfield component in spacer https://github.com/floored/workplace/blob/master/assets/js/spacerPro/components/users/form/InviteUserForm.jsx#L55-L60

note that the state is managed by our spacer pro file, we just pass in the updated value

hwigmore:
for the preview examples, you'll just need to set an initial state and make the update happen to that state in `TextFieldPreview.jsx` in the way other components (like the conditional suffix one and validity one) do

joshkarges:
It seemed a little redundant to force every `TextField` component to have both a `value: {this.state.myValue}` and an `onChange: {(e) => this.setState({ myValue: e.target.value })`.  So I tried to make that more implicit to the component.

You should be able to use your component in spacer the same way (the "Conditional Suffix Icon" example and the "Validation" example both keep track of the input text state outside of the TextField component).  

This would just allow someone to have a TextField component without specifying `value` and `onChange`.

hwigmore:
oh interesting point. thinking about it

emma-floored:
what's the use case (besides previews on the blocks documentation) where we would have a text field and not want to specify a `value` or `onChange`?

joshkarges:
Hmmm, the use cases are probably not plentiful, but I'll try for one:

If I have a search bar, where I don't need to keep track of the actual search query, I just need to send it into my `performSearch` handler, but I don't want to perform the search on every key stroke, just on Enter (because maybe its expensive).
```
<TextField label="Search" onKeyDown={(e)=> (e.key === 'Enter') && performSearch(e.target.value)} />
```

joshkarges:
Also, Heather, with this change you wouldn't need to add the `value` prop on your Spacer code, unless you wanted to show a different starting text in the box than `''`.  So since your `state.firstName` already starts as `''`, you won't have to add the `value` prop there.  Yay deleted code!

emma-floored:
if we default value to null instead of empty string it fixes a few of the text box previews (but not the one that has a value already, since that one doesn't have an onChange I think?)

joshkarges:
interesting!

joshkarges:
Re: defaulting `value: null`
It seems like that "Filled" example just doesn't change because the specified `value` is a constant string instead of a state tracked and set variable.  The other examples with a `value` prop change because they track and set them correctly in the `value` and `onchange` props.

joshkarges:
I understand the worries about both A.) keeping track of a state in 2 places, and B.) not having consistent mechanics between other components.

joshkarges:
Hey!  Sorry, I left this for so long, things got busy on the Plans side.

So, ultimately, this issue of requiring the `value` prop is up to the people who use React.  It seems my opinion is a bit out numbered here, so I can revert these changes and just try to fix the text cases without changing the component.

We have found use cases on the Plans side where we don't want to externally track the `value` of a textField.   One example, is in the Furniture Unit Input Form, where we want to take the input of the textField and immediately create a new manifest type with it `onchange` (so it would be useless to us to have to have to store this value externally somewhere).

This is what our component looks like:
```
m(textField, {
  placeholder: 'Add furniture piece label',
  onchange: ctrl.editManifestType(-1)
})
```

And here's how it behaves:
uncontrolledtextfield

I'll still continue to revert the changes, I just though you'd be curious about how it's working for Plans.

hwigmore:
i dont totally get it. don't you need the value you typed into the text field? where/how do you pass that back to do something with it?

joshkarges:
The value of the text field is only used for that `onchange` handler.  The onchange handler then uses that value to create a new component, but it doesn't need to feed it back in to the 'Add furniture piece label' component.

hwigmore:
oh okay. i was confused because i was expecting your example onchange to take in evt.target.value or something but it took in a -1 instead

joshkarges:
ah yeah, the `ctrl.editManifestType(-1);` returns a handler which takes the `evt` as input.

joshkarges:
Changed this PR to only fix the textField docs previews (keeping the title of the PR the same, because I've heard there's annoyances with email threading).  Let me know if you'd like any other changes.

hwigmore:
i'm having trouble running blocks locally but this looks good so 👍

PR #56, opened 2018-09-24
add vertical ellipsis for popover

hwigmore:
@emma-floored @MagdalenaHenke 

screen shot 2018-09-24 at 7 28 07 pm


hwigmore:
https://app.clubhouse.io/floored/story/10772/add-vertical-ellipsis-icon-for-popover

hwigmore:
trying to figure out why ModalPreview and ButtonPreview changed as well...

hwigmore:
oh looks like `ButtonPreview.js` and `ModalPreview.js` had changes that never got checked in (on master if i run `gulp server` i also see those files have been updated). i'll just leave those updates here

hwigmore:
i added one more thing in here 🙊 

in our use case we need to `evt.preventDefault()` when we click our popover menu (it is within a larger button/link). i think/hope adding this on the blocks level will be ok for other popover use cases?

emma-floored:
I think the `evt.preventDefault()` should be fine... looks good! 👍

PR #57, opened 2018-09-27
add modal react previews in patterns section

hwigmore:
[Clubhouse](https://app.clubhouse.io/floored/story/10823/add-react-modal-previews-to-documentation)

@emma-floored i know these haven't been officially spec'd out yet, but since they exist and since we use them, it felt like we needed a react preview as a start. 

liz thought Patterns was the right place for them


react/modals/ActionModal.jsx
```diff
@@ -7,6 +7,8 @@ const ButtonSecondary = require('../buttons/ButtonSecondary.jsx');
 
 
 const ActionModal = (props) => {
+  if (props.isHidden) return null;
```
hwigmore: `BaseModal` has `props.children` as required, so we were getting an error how this was (making `modalContent = null`). i figured this is clearer anyway and saves a few steps below!

docs/patterns/modals.html
```diff
@@ -0,0 +1,15 @@
+---
+layout: componentSection
+section: Components
```
emma-floored: change to "Patterns"?

emma-floored: oh actually no, the Tabs page also still says "Components"... basically when I run this, the modals link doesn't show up under "Patterns", and I'm trying to figure out why

emma-floored: But it does seem to be fixed if you change this to "Patterns". But if you update it, could you also update it on tabs.html ?? 💐 💐 💐

emma-floored:
other than my one series of comments, this looks good! 👍

docs/patterns/modals.html
hwigmore: oh i copied and pasted this file from tabs yeah! will update both!

PR #61, opened 2018-10-05
V1 updates

emma-floored:
@hwigmore @Jinsoo94 this finishes up the tickets that were associated with V1 I think? So I guess after this, we could call on the designers to review.

https://app.clubhouse.io/floored/story/10096/update-alerts
https://app.clubhouse.io/floored/story/10099/update-dropdowns
https://app.clubhouse.io/floored/story/10104/update-text-fields

Updates:
- Alerts
  - updated close icon styles
  - decided to leave alert positioning to individual apps to implement, as well as grouped alerts
- Dropdowns
  - added option for dropdown with an icon
  - added spacing between consecutive vertical dropdowns
    - ** in hindsight, I'm gonna test this some more, seems like the way I did it would mess up horizontally aligned dropdowns...
  - keyboard interactions
    - @Jinsoo94 mind checking this out to make sure it's ok? I added:
      - pressing a character key should select the first option that starts with that character
      - tab doesn't cycle through the options, it closes the menu and goes to the next input element
- Text Fields
  - add date/time picker options
  - allow icon next to text field
  - add option for multiline text area

screen shot 2018-10-04 at 9 12 30 pm
screen shot 2018-10-04 at 9 13 12 pm
screen shot 2018-10-04 at 9 13 17 pm
screen shot 2018-10-04 at 9 13 22 pm


emma-floored:
* I notice I messed up the overflow dropdown example, will fix tomorrow

hwigmore:
re: "tab doesn't cycle through the options, it closes the menu and goes to the next input element" is that the tab behavior for standard select dropdowns?

blocks-styles/alerts.styl
```diff
@@ -14,10 +14,23 @@
...
+    &:hover {
+      background-color: $interactive-hover;
+    }
+    &:focus {
+      background-color: $selection;
+    }
   }
 }
 
+.blx-alerts-container {
```
hwigmore: not actually blocks level, per discussion in slack?

blocks-styles/dropdowns.styl
```diff
@@ -1,5 +1,9 @@
 .blx-dropdown-wrapper {
   display: inline-flex;
+  
+  & + .blx-dropdown-wrapper {
```
hwigmore: you mentioned that this would mess up two horizontal dropdown siblings. can remove? seems like an app-level tweak if you want to style multiple in relation to each other?

docs/components/dropdowns.html
```diff
@@ -115,6 +162,7 @@
 <!-- state & behavior section -->
 <div class="l-flex-vertical doc-section">
   <h4>States & Behavior</h4>
+  <p>*Dropdowns require JavaScript to work properly - for working examples, see the "React Components" section below.</p>
```
hwigmore: 👏

react/input/TextField.jsx
```diff
@@ -74,6 +65,84 @@ class TextField extends React.Component {
...
+
+    // icon element
+    let iconElement = null;
+    if (this.props.icon) {
+      iconElement = <span className="blx-text-field-icon">{this.props.icon}</span>;
+    }
+
+    let textInputElement = null;
+    if (this.props.isMultiline) {
+      textInputElement = this.getTextArea();
```
hwigmore: i would vote to make `TextArea.jsx` its own component, and not a variation of of `TextField.jsx`, since. i realize there would be a lot of overlap. but i think it would be worth it for cleaner/shorter files. what do you think?

hwigmore:
this might be a question for design, but i think the calendar icon should be `$interactive` color because it currently looks disabled

Jinsoo94:
@emma-floored Just tested dropdowns. They are working properly!

emma-floored:
@hwigmore re: behavior for standard select dropdowns

Just checked, and hitting tab when a standard select dropdown is open has no effect. I played around, and this does seem to be the only difference between the two.

emma-floored:
@hwigmore @Jinsoo94 pushed a fix for a bug with the dropdown menu where `onChange` was called multiple times due to event listeners being reregistered.

react/input/TextField.jsx
```diff
@@ -74,6 +51,48 @@ class TextField extends React.Component {
...
     );
+  }
+
+  getTextArea() {
+    let textAreaRef = null;
+    if (this.props.forwardedRef) {
+      textAreaRef = this.props.forwardedRef;
+    } else {
+      textAreaRef = this.textAreaRef;
+    }
```
Jinsoo94: what does getTextArea() do in this file? Seems like it is not being called anywhere else. + Its missing a closing } !

emma-floored: good call... it should be removed, since I split them into 2 files

react/input/TextArea.jsx
```diff
@@ -0,0 +1,133 @@
...
+  updateValidity() {
+    this.setState({
+      isValid: this.props.isValid(this.props.value)
+    });
+  }
+
+  resizeTextArea() {
+    if (!this.textAreaRef) return;
+    this.textAreaRef.current.style.height = '1px';
+    this.textAreaRef.current.style.height = `${this.textAreaRef.current.scrollHeight}px`;
```
Jinsoo94: Just wondering, why does height have to set to 1px before being set to scrollHeight?

emma-floored: ooh I should comment this... it's so that `scrollHeight` will be the entire height needed to fit all the text.

hwigmore:
this might be a silly question, but is all the resizing stuff necessary? doesn't `textarea` have that built in or no?

hwigmore:
also can we make `rows` a prop that has a default (4 maybe?) which necessarily sets the min/starting height? 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea

emma-floored:
@hwigmore textarea does not have that stuff built in - it will have a scrollbar

re: a `rows` prop - is that only useful if you don't want the `textarea` to resize to fit whatever content is in it?

hwigmore:
ooh interesting. maybe (re: `rows`). ignore in that case!

PR #62, opened 2018-10-08
apply blx-invalid class to text field input

hwigmore:
@emma-floored small bug i noticed! the error message was showing up but not the red outline


emma-floored:
👍

PR #63, opened 2018-10-08
give validity power back to the owner

hwigmore:
https://app.clubhouse.io/floored/story/11089/textfield-give-validity-check-power-to-owner-caller-of-the-component

discussed this with @emma-floored but: spacer pro needs to do its own validity check at times other than `onKeyDown` or `onBlur`. this PR makes it so that `TextField.jsx` just takes an `isValid` boolean instead of function, so the owner can choose when to call it (that might be `onKeyDown`, `onBlur`, both of those, and/or something else entirely, like when the user hits a form submit button)

this also means that `TextField.jsx` no longer needs state and can be a pure component

in the react preview example here, i made the validity check happen `onChange`. one weird thing about this is if you alter the text in the field, you'll get the error message but still see the focused (green) border. should we have the error border override the focus border?
- update: i posed this question in the #blocks channel
- update to the update: there's a clubhouse ticket for this now https://app.clubhouse.io/floored/story/11251/error-outline-red-should-take-precedence-over-focus-outline-green-if-text-field-value-is-invalid

@emma-floored @MagdalenaHenke @Jinsoo94


hwigmore:
yikes conflicts. fixing them up. dont look at me!

hwigmore:
ok rebased and fixed conflicts 🤞 i didnt ruin anything

can someone take a look at this today since it's blocking spacer pro stuff?

hwigmore:
oh i'll also tag @tariqporter in case you want to take a look!

react/input/TextField.jsx
```diff
@@ -2,99 +2,71 @@ const React = require('react');
...
-    );
-  }
+      <div className="blx-text-field-container">
+        {prefixElement}
+        {suffixElement}
+        <input
+          className={textFieldClasses}
+          type={props.type}
+          name={props.name}
+          value={props.value}
```
hwigmore: one thing i changed here is on master it was `props.value || ''` but that fallback shouldn't be necessary since we define it in `defaultProps`

react/input/TextField.jsx
```diff
@@ -2,99 +2,71 @@ const React = require('react');
 const PropTypes = require('prop-types');
 
 
-class TextField extends React.Component {
-  constructor(props) {
-    super(props);
+function TextField (props) {
```
Jinsoo94: Not a huge problem, but it seems like other pure components use arrow function syntax. I think it would be better to stick to one style!

hwigmore: oops, good eye!

emma-floored:
👍

PR #67, opened 2018-10-12
error border should override focus/success border

hwigmore:
[Clubhouse](https://app.clubhouse.io/floored/story/11251/error-outline-red-should-take-precedence-over-focus-outline-green-if-text-field-value-is-invalid)

this is the thing that runi gave a "soft yes" on, so figured we can try it out

this PR is based on the error-asterisk PR to avoid conflicts


docs/_javascript/TextFieldPreview.jsx
```diff
@@ -7,10 +7,10 @@ const TextField = require('../../react/input/TextField.jsx');
 const TextArea = require('../../react/input/TextArea.jsx');
 const TextFieldWithDropdown = require('../../react/input/TextFieldWithDropdown.jsx');
 
-const INVALID_VALUE = 'Wrong';
-
+const INITIAL_INVALID_VALUE = '';
+const VALID_VALUE_MIN_LENTH = 5;
```
hwigmore: i updated this preview example because the previous example of "Wrong" only had that exact window during which to show the error outline, but now you can type a few letters and watch it change

blocks-styles/text-fields.styl
```diff
@@ -78,6 +78,10 @@
     &:focus, &.blx-focus {
       outline: none;
       border: 1px solid $success;
+
+      &.blx-invalid {
```
emma-floored: instead of nesting this like this, does it work if you just move the `.blx-invalid` class that is above this `focus` block to below it instead?

hwigmore: oh good call, will try!

hwigmore: updated!

emma-floored:
👍

PR #68, opened 2018-10-18
fix dropdown bugs

emma-floored:
@Jinsoo94 @hwigmore this takes care of 2 issues

- since we were attaching event listeners on the dropdown's `componentDidMount`, conditionally added dropdown menu items weren't getting event listeners properly attached
  - I pulled out the menu items into their own component - `DropdownItem` - and, since we no longer need to consider arbitrary elements (from @hwigmore) we don't have to attach event listeners that way anymore - I just use the regular `onKeyDown`/`onKeyUp`/`onClick` for each item
- I noticed that popover menus weren't controllable via keyboard like the dropdown menu was
  - I ended up combining PopoverMenu and Dropdown... the result is pretty ugly, but I wasn't sure what a better solution would look like. Definitely open to suggestions here. My original idea was to separate out the keyboard functionality into a separate component or wrapper or something (like closeOnClick), but I wasn't able to get that working how I wanted (it would have to include the trigger button and the menu, since you can open the menu with down arrow, and at that point... it's the whole menu).


Jinsoo94:
Nice! I like the new component DropdownItem + I like the pattern where Dropdown serves as a basis for variational components like PopoverMenu. I don't feel too strongly about isPopover prop being part of Dropdown.jsx; I think it would make reusing Dropdown component harder in the future as we try to use it to make another dropdown-ish component. That would require prop like isSomeOtherDropdown, and more conditionals to control the className/trigger type which would end up crowding Dropdown.jsx. 

Maybe getTrigger() => getDefaultTrigger() and remove `if (this.props.isPopover) { ... }`, and in render change `{this.getTrigger()}` to `{this.props.trigger ? this.props.trigger : this.getDefaultTrigger()}`? (trigger would be defined in the client component and be passed down) 
`{this.getPopoverText()}` can be handled similarly. **BUT** I am unsure what the cleanest way would be to handle classNames :-p

emma-floored:
@Jinsoo94 I refactored and was able to pull the interactions out into a wrapper - mind taking another look?

hwigmore:
sooo i hadn't looked at this yet since jinsoo made that first comment which i knew you were gonna address.

but thennnnn lena mentioned a similar refactor as this to me today. so i'm thinking...can i pass my review off to her? 😈 (i already asked her)

hwigmore:
@MagdalenaHenke

hwigmore:
@MagdalenaHenke if you dont have time to get to this, i'll review it. but i figured i'd throw your name in the ring in case you can get to it before me!

Jinsoo94:
Also looking at it again 👀

Jinsoo94:
👍

react/dropdowns/Dropdown.jsx
```diff
@@ -1,224 +1,83 @@
...
+const DropdownMenu = (props) => {
+  let content = props.text;
+  let triggerClassNames = 'blx-dropdown-trigger';
+
+  if (props.isOpen) triggerClassNames += ' blx-active';
+  if (props.isDisabled) triggerClassNames += ' blx-disabled';
+  if (props.value) triggerClassNames += ' blx-dropdown-trigger-placeholder';
+
+  if (props.value) {
+    for (let i = 0; i < props.options.length; i++) {
```
hwigmore: this might be personal pref so feel free to ignore, and i know this wasn't new to this PR, but i have a slight preference to simplify this loop to just two lines for readability
```
if (props.value) {
  const selectedOption = _.find(props.option, ['value', props.value]);
  let content = selectedOption.text || null; 
}
```
or, might not even need that fallback (null vs undefined, would that make a difference here?) and can do it in one line
`let content = _.find(props.option, ['value', props.value]).text;`

react/dropdowns/Dropdown.jsx
```diff
@@ -1,224 +1,83 @@
...
-                    {item}
-                  </li>
-                );
-              })
-            }
-          </ul>
-        </div>
+        }
+        { trigger }
+        <ul className={`blx-dropdown-menu ${props.isOpen ? '' : 'blx-hidden'}`}>
```
hwigmore: `className={menuClasses}` (or, can delete lines 47-48)

react/dropdowns/PopoverMenu.jsx
```diff
@@ -1,85 +1,94 @@
...
+  text: null,
+  popoverIcon: 'more-horizontal',
+  value: null,
+  onChange: () => {},
+  autoFocus: false
 };
 
 
-module.exports = closeOnClick(PopoverMenu);
+module.exports = closeOnClick(keyControlledMenu(PopoverMenu));
```
hwigmore: so fancy!

hwigmore:
from testing: 

1. once i select a dropdown item, it still has the placeholder look
master / branch
screen shot 2018-10-25 at 11 11 43 am screen shot 2018-10-25 at 11 11 34 am


2/ the react previews for popovers are all the same

master / branch
screen shot 2018-10-25 at 11 14 08 amscreen shot 2018-10-25 at 11 13 58 am

react/dropdowns/Dropdown.jsx
emma-floored: currently I don't have lodash included in Blocks... but I can see if I can simplify this

hwigmore: oh no need to add it just for this. wow, i'm impressed that it's not used anywhere though! 👏

PR #72, opened 2018-10-25
don't crash dropdowns with dynamic options when using keyboard navigation

MagdalenaHenke:
[Clubhouse](https://app.clubhouse.io/floored/story/11841/dropdowns-with-dynamically-created-fields-crash-when-using-keyboard-arrows)

For dropdowns with dynamically created content and changing options, I noticed that the keyboard navigation was crashing them. That was because we never created the optionsRefs if the options changed over time. 

I think this fixes it. @emma-floored - would you have a look? (or happily anyone else who'd like to review)


emma-floored:
omg thank you lena 💐 👍

PR #73, opened 2018-10-29
vertically align calendar icon on textfield

Jinsoo94:
vertically align calendar icon on date textfield regardless of the height of the input
@emma-floored


emma-floored:
@Jinsoo94 rather than just modifying this on the final generated `blocks.css` file, you should modify it here https://github.com/floored/blocks/blob/master/blocks-styles/text-fields.styl#L34 and regenerate `blocks.css` by running `gulp server`

emma-floored:
👍

PR #75, opened 2018-11-01
FIX: define error-hover-hex color

MagdalenaHenke:
@emma-floored @hwigmore 

Our staging styles are broken, and the heroku log pointed me at `error-hover-hex` being undefined. I think it _is_ undefined.  

I wasn't sure where to look, I just took the `danger-active` color for `error-hover-hex`. What should it be?
screen shot 2018-10-31 at 8 18 11 pm


hwigmore:
it's possible that we don't even need / use the root level `error-hover` at all and can delete the reference to it rather than adding this. because when we hover on an error button we just invert (go from red background white text to white background red text)

@emma-floored can you confirm?

MagdalenaHenke:
Oooh, I think Emma is out for a little while. I'd trust you if you think that's right. Or @Jinsoo94 ?

hwigmore:
i'm having local issues running blocks, but i'm fairly confident that the fix to this is to just remove the reference to `error-hover-hex` from `variables-base-root.styl` and rerun gulp (will leave 3 files updated)

hwigmore:
i think this is fine to merge and pull into spacer and see if it fixes staging 👍

MagdalenaHenke:
K. Merging. I pulled most recent datavis and global searched for `error-hover` and nothing came up so at least it won't break stuff there.

PR #79, opened 2018-11-16
if we always require activeTabIdx, add state to react preview components for RadioTabs

MagdalenaHenke:
Bit of a journey:

On current master, this is broken (which I noticed in Spacer where we are now using RadioTabs):
radiotabblocks

I traced it back to this commit: https://github.com/floored/blocks/pull/76/files

The main thing in that commit is the change from `defaultChecked={idx === props.activeTabIdx}` to `checked`. 
When using `defaultChecked`, there was some css magic that I'm honestly not fully understanding that made the highlighting of the most recently clicked tab work. The change to `checked` requires everyone to pass in the `activeTabIdx` prop to get any tab to highlight. 

a) I want to make sure losing that css magic was on purpose and what we want. @emma-floored ?
b) If that _is_ what we want, I think it makes sense to both 
- have the default `activeTabIdx` be `null` instead of 0, cause we don't know whether the first tab is always active on page load.
- to update the React Preview components to actually have state and update. So I added that. If we require passing in `activeTabIdx`, I figure it's only fair to pass up `idx` for a given tab in the `onChange` function as well so the parent component can keep track of that in their own state. 

@Jinsoo94 , also tagging you since that change came from you.


MagdalenaHenke:
https://app.clubhouse.io/floored/story/12356/radio-tabs-selection-broken

docs/_javascript/TabPreview.jsx
```diff
@@ -6,54 +6,70 @@ const getPreviewComponent = require('./common/getPreviewComponent.jsx');
...
-              'Tab 1', 'Tab 2', 'Tab 3', 'Tab 4'
-            ]
-          }
-          isVertical
-          onChange={() => {}}
-        />
-    ) }
-  </div>
-);
+class TabPreview extends React.Component {
```
MagdalenaHenke: Definitely look at this file **without** the whitespace changes! Most of this is the same, it's just indented

react/tabs/RadioTabs.jsx
```diff
@@ -16,7 +16,7 @@ const RadioTabs = (props) => {
               value={tab}
               checked={idx === props.activeTabIdx}
               disabled={props.isDisabled}
-              onChange={props.onChange}
+              onChange={evt => props.onChange(evt, idx)}
```
MagdalenaHenke: figure it's just fair to pass the `idx` up as well if we require the parent component to pass back the `activeTabIdx`

docs/_javascript/TabPreview.jsx
```diff
@@ -6,54 +6,70 @@ const getPreviewComponent = require('./common/getPreviewComponent.jsx');
...
+                ]
+              }
+            />
+        ) }
+        { getPreviewComponent('Vertical',
+            <RadioTabs
+              tabs={
+                [
+                  'Tab 1', 'Tab 2', 'Tab 3', 'Tab 4'
+                ]
```
MagdalenaHenke: Ignoring the LinkTabs above because that's a little bit more effort than just state (I mean I figure those will be based on url....but also, maybe those should be React Links so we don't lose state?) and wasn't feeling that.

hwigmore:
this makes sense to me! 👍 

@Jinsoo94 or @emma-floored any thoughts?

emma-floored:
Oops!

This makes sense to me 👍

PR #80, opened 2018-11-19
fix sortable list behavior

emma-floored:
@Jinsoo94 

- sortable lists should work in firefox now
- sortable lists shouldn't break so easily
- added lines to show where the item will drop
- can now drag over whole list item rather than having to hit the drag icon

I added sort of a fake preview page at http://localhost:8080/patterns/sortablelists/ (can't access by navigation) with a React component so you can test!

### DIFF
--- a/blocks-styles/sortable-list.styl
+++ b/blocks-styles/sortable-list.styl
@@ -3,26 +3,32 @@
   list-style: none;
   margin: 0;
   padding: 0;
-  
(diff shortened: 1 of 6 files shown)


Jinsoo94:
👏 👏 👏 wow code is much cleaner than I thought it would be! + test works fine (actually a lot smoother than before)

PR #82, opened 2018-11-30
updated typography according to the new spec

Jinsoo94:
This branch updates typography based on the new spec! @emma-floored @hwigmore

### DIFF
--- a/blocks-styles/typography.styl
+++ b/blocks-styles/typography.styl
@@ -2,9 +2,10 @@
 blxBaseText() {
   font-weight: $normal-weight;
   font-size: $base-font-size;
-  line-height: 28px;
+  line-height: 24px;
   text-align: left;
   color: $text-01;
+  margin-bottom: 12px;
 }
 
 blxLink() {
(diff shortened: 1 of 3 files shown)


Jinsoo94:
https://app.clubhouse.io/floored/story/12532/update-typography

emma-floored:
Awesome! Would you mind fixing one thing for me? The `blx-subtitle` on a dark background isn't quite what we want, I think. I made the background part of the style, but I think actually we just want like a `blx-subtitle light` or something that has the light text style that we can put over any background...

Jinsoo94:
@emma-floored updated!

emma-floored:
thanks!! 👍

blocks-styles/typography.styl
```diff
@@ -94,6 +124,10 @@ a, .blx-link {
   blxLink()
 }
 
+h1.blx-XL {
+  blxHXL()
+}
+
```
hwigmore: this is really silly and picky, but i feel like it's more clear to read and find things when the generic stuff (regular h1) comes first, so
```
h1 {
  blxH1()
}
h1.blx-XL {
  blxHXL()
}
```
(or could nest it with an `&.blx-XL`

hwigmore: ooh i see why you did it this way. so they are ordered by size. fine to leave as is

Jinsoo94: is it good to go?

PR #83, opened 2018-12-05
add invalid option for dropdowns

hwigmore:
@emma-floored i added the option to pass an `isValid` prop to dropdowns in the same way we do to text fields (mentioned on #blocks channel).

i added an example on the docs but not a separate react preview for it, because it didn't seem like it needed a whole separate interactive preview just for that style variation. thoughts?


hwigmore:
https://app.clubhouse.io/floored/story/12678/add-invalid-border-option-for-dropdowns

hwigmore:
oh one thing i noticed while working on this branch is that things jump around a little when we hover on a dropdown. will fix that in a separate pr

hwigmore:
@MagdalenaHenke mind taking a look?

hwigmore:
@helenCBRE

hwigmore:
rebased and f-pushed!

PR #85, opened 2018-12-06
stopPropagation on closeOnClick toggle

helenCBRE:
added stopPropagation to toggle in closeOnClick to prevent click on element that closeOnClick element is on top of

### DIFF
--- a/react/wrappers/closeOnClick.jsx
+++ b/react/wrappers/closeOnClick.jsx
@@ -31,6 +31,7 @@ function closeOnClick(WrappedComponent) {
     }
 
     toggle(e) {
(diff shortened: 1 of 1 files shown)


helenCBRE:
@hwigmore can you take a quick peek at this? (is this the right pr process for blocks btw?)

hwigmore:
tested dropdowns and popovers and both seem good! 👍 

i have two open prs in blocks. would you mind testing those as well? and then i also just noticed something small that is off (alignment in popovers) that i'd like to fix. and then maybe today we can update with all 4 little tweaks, to go from v0.1.0 to v0.1.1

PR #86, opened 2018-12-06
update package.json and readme for versioning

hwigmore:
@emma-floored @hamham91


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "blocks",
-  "version": "1.0.0",
+  "version": "0.1.1",
```
hwigmore: not sure if we want/need to manually update this after every release?

hamham91: let's do that!

hamham91:
👍

PR #87, opened 2018-12-07
popover vertical alignment and subtitle doc updates

hwigmore:
[Clubhouse](https://app.clubhouse.io/floored/story/12801/fix-popover-text-icon-vertical-alignment)

the first commit isn't directly related to the bug, but i think it's helpful to see that that class does come with a built-in margin


Jinsoo94:
Looks great! 

One thing, this is super nitpicky + my eye could just be weird, so I want more people's input, but I think blx-icon is not at the vertical center of blx-popover-trigger. Adding `display: flex` to .blx-popover-trigger vertically aligns it.

hwigmore:
ooh good eye! will update

Jinsoo94:
👍

emma-floored:
👍

PR #88, opened 2018-12-07
text validity updates

hwigmore:
[Clubhouse](https://app.clubhouse.io/floored/story/11286/textarea-give-validity-check-power-back-to-the-owner-caller-of-the-component)

context for this: previously i made made this change (#63) so that the owner of the textfield component can choose when to call its validity function, and just let the blocks component know the result (maybe the owner wants to call it on keydown and blur, but maybe they don't, and maybe they wanna call it on a form submit or something).

this pr is to make that same update on `TextArea`


react/input/TextArea.jsx
```diff
@@ -6,23 +6,11 @@ class TextArea extends React.Component {
...
 
-    this.state = {
-      isValid: props.isValid(props.value)
-    };
-
     this.textAreaRef = React.createRef();
   }
 
   componentDidMount() {
-    if (this.props.isMultiline) {
```
hwigmore: unrelated: removed the `isMultiline` prop, which i think was a relic of of textfield and textarea sharing a component. resize now gets called on mount

react component preview on master
screen shot 2018-12-07 at 11 57 06 am



on branch
screen shot 2018-12-07 at 11 56 48 am

react/input/TextArea.jsx
```diff
@@ -34,46 +22,14 @@ class TextArea extends React.Component {
     this.textAreaRef.current.style.height = `${this.textAreaRef.current.scrollHeight}px`;
   }
 
-  getTextArea() {
-    return (
-      <textarea
```
hwigmore: just moved this into the `render` function to simplify

react/input/TextFieldWithDropdown.jsx
```diff
@@ -22,7 +22,7 @@ TextFieldWithDropdown.propTypes = {
     placeholder: PropTypes.string,
     invalidErrorMessage: PropTypes.string,
     value: PropTypes.string,
-    isValid: PropTypes.func
+    isValid: PropTypes.bool
```
hwigmore: updating this to reflect the  `TextField` component it points to

react/input/TextArea.jsx
Jinsoo94: hmmm for some reason I cant see the height change in the deploy preview. It looks like the master.

Jinsoo94:
In the preview, the textfield with dropdown alignment is off like this!
screen shot 2018-12-07 at 12 52 45 pm

react/input/TextArea.jsx
hwigmore: are you looking at the react component or the plain html example? i did the same thing but then i scrolled down and found it!

hwigmore:
good catch! since that's on master and out of scope i'll add a bug ticket for it!

react/input/TextArea.jsx
Jinsoo94: ohhh you are right. React component looks correct. Just HTML example and edge case example seem not to change height.

hwigmore:
@Jinsoo94 any other thoughts?

Jinsoo94:
Oh oops forgot give a thumbsup. All good! 👍

hwigmore:
@Jinsoo94 just an FYI that my most recent pr fixed this

Jinsoo94:
@hwigmore ok just checked it 👍

PR #90, opened 2018-12-07
add error message option for dropdown

hwigmore:
[Clubhouse](https://app.clubhouse.io/floored/story/12842/allow-dropdown-to-have-error-message)

Impetus: 
- invalid dropdowns should display an error message, if one is passed

What else this triggered:
- creating a shared `blx-invalid-input-message` class, in place of the previous one that was named text field
- some refactoring of the dropdown component just because i was having a little trouble with all the classes that were so far from where they were used
- I made the invalid error message absolutely positioned, so that things don't get pushed around below the dropdown/input when it comes and goes


blocks-styles/dropdowns.styl
```diff
@@ -16,15 +16,14 @@
...
-  margin: 0;
   cursor: pointer;
   white-space: nowrap;
   overflow: hidden;
   text-overflow: ellipsis;
-
   background-color: $ui-00;
   border: 1px solid $ui-03;
   blxUIPlaceholder();
+  margin: 0;
```
hwigmore: moved this down from line 19 because we want it to override the margin-bottom that gets set in the mixin above

blocks-styles/dropdowns.styl
```diff
@@ -67,21 +66,8 @@
...
-
-    .blx-icon {
-      margin-top: 11px;
-
-      & + span {
-        margin-top: 9px;
-      }
-    }
+    border: 1px solid $ui-03;
+    box-shadow: 0 0 0 1px $ui-03;
```
hwigmore: i noticed that there was some weird hover shifting happening on an error dropdown (https://app.clubhouse.io/floored/story/12842/allow-dropdown-to-have-error-message) which i had missed before. that got me looking at this whole section and i realized if i replaced it with what we have for text field hover, the problem goes away and all the other (not invalid) dropdown hovers still look the same!

blocks-styles/text-fields.styl
```diff
@@ -5,6 +5,7 @@
 
 .blx-text-field {
   display: inline-flex;
+  position: relative;
```
hwigmore: to allow for absolute positioning of the invalid message

blocks-styles/toggles.styl
```diff
@@ -42,7 +42,7 @@
         color: $ui-00;
       }
     }
-    
+
```
hwigmore: whaat didnt realize i was in this file

blocks-styles/typography.styl
```diff
@@ -206,6 +206,14 @@ h6 {
   margin-bottom: 6px;
 }
 
+.blx-invalid-input-message {
+  color: $error;
+  font-size: 12px;
+  line-height: 16px;
+  position: absolute;
+  top: 100%
+}
```
hwigmore: including positioning felt a little sketchy. but this also seemed like the best file for this shared class that gets used in both text-fields.styl and dropdowns.styl

helenCBRE:
@hwigmore hey i took a look at the code, will take a closer look in a bit. It seems that the branch hasn't deployed so I don't know how to look at these changes on a staging link. Do you have access on heroku or aws to deploy it? Or is that not how stuff works for blocks?

hwigmore:
ah it did deploy but it's not really obvious. if you click on the "details" link on the deploy/netlify, it'll take you there! 

as a reminder, you'll see the update in the React components section, not the regular style section so keep scrolling!

helenCBRE:
should users be able to go back to a state where nothing is selected in the dropdown and they see the invalid state warning again?

hwigmore:
in this case, no, because it's only invalid with a placeholder and once you select something the placeholder disappears! i think that is desired behavior

blocks-styles/typography.styl
helenCBRE: why is including positioning sketchy?

docs/_javascript/DropdownPreview.jsx
```diff
@@ -86,6 +87,24 @@ class DropdownPreview extends React.Component {
               />
           ) }
         </div>
+        <div className="l-flex-horizontal">
+          { getPreviewComponent('Invalid',
+              <Dropdown
+                text="Placeholder"
+                isValid={!!this.state.value4}
```
helenCBRE: why is validity based on `value4` and not all the options?

docs/_javascript/DropdownPreview.jsx
```diff
@@ -86,6 +87,24 @@ class DropdownPreview extends React.Component {
...
+                invalidErrorMessage="Please pick an option"
+                onChange={evt => this.setState({ value4: evt.value })}
```
emma-floored: this seems to not be working? I can make selections in all the other dropdowns, but not in this one...

hwigmore: oops that's bc i forgot `value` on here

hwigmore: should be fixed on last commit

hwigmore:
rebased and force pushed to fix a conflict

emma-floored:
👍

PR #91, opened 2018-12-10
TextField and TextArea: do not render empty div if no label, remove old invalidLabelMarker

hwigmore:
really minor clean up, but i noticed that we were adding empty `<div>`s to the dom when the text field has no label. in the process i also discovered some residual stuff from the asterisk we used to show for invalidity

@emma-floored


hwigmore:
@tariqporter would you mind taking a look at this? would love to get it in with your other TextField change for the next release

emma-floored:
👍

emma-floored:
I did notice one weird thing but it's on master as well, so wasn't caused by this branch...

These side by side text fields (one with label, one without) don't line up.
screen shot 2018-12-18 at 3 46 03 pm

hwigmore:
oh yeah in spacer pro when we want that to happen we pass in an empty space string for the label on one. but maybe it'd make more sense to have a `leaveSpaceForLabel` boolean prop

tariqporter:
I’ll take a look!

hwigmore:
it's okay i got a thumbs up from emma! thank you though!

PR #92, opened 2018-12-17
Remove blur() in TextField when pressing Enter

tariqporter:
@hwigmore @hamham91 

Very minor change to alter default behavior. 
`blur` prevents us focusing a particular element on `Enter` keypress, as `blur` essentially focuses `body`.


emma-floored:
👍

hwigmore:
minor, could simplify to match all the other event handlers now (`onKeyUp={props.onKeyUp}`) but either way 👍

tariqporter:
@hwigmore Good catch

hamham91:
👍

hamham91:
Is there a case where we would want this behavior enabled via a flag?
```
<TextField
   ...
   blurOnEnter
/>
```

tariqporter:
> Is there a case where we would want this behavior enabled via a flag?
> 
> ```
> <TextField
>    ...
>    blurOnEnter
> />
> ```

I think in the case that is wanted behaviour, the user should put `blur` in their own function

hwigmore:
i'm open to a flag, but i also think that if a particular app wants that behavior, they can also just pass it as part of their `onKeyUp` or `onKeyDown` function.

hwigmore:
👍 to merge! (or squash and merge, as we're starting to do for blocks)

can you also open up a new release called v0.1.3 and describe this change? in the same format as the other releases? 

thanks!!

hwigmore:
gonna squash and merge this for you and add to the release notes...

tariqporter:
Thanks!

PR #97, opened 2019-01-16
Update to Radio Buttons

JanCBRE:
#### New stuff:
- Invalid state for individual radio buttons and radio button groups
- Active state for radio buttons

#### Minor Changes:
- Radio button size and spacing 
- Radio button background color

I updated the radio button preview to include the additions.


emma-floored:
https://app.clubhouse.io/floored/story/13297/update-radio-buttons

react/input/RadioButtonGroup.jsx
```diff
@@ -9,13 +9,14 @@ const RadioButtonGroup = props => (
...
+          key={option.text}
           name={props.name}
-          text={option}
-          value={option}
+          text={option.text}
+          value={option.text}
           isDisabled={props.isDisabled}
           onChange={props.onChange}
           isChecked={idx === props.checkedIndex}
+          isValid={option.isValid === false ? option.isValid : props.isValid}
```
hwigmore: so `props.isValid` means that the entire radio group is valid or not? 
but `option.isValid` is just for that specific radio button?

did you find out from design what cases they are imagining?

JanCBRE: Yeah. To quote Jürgen: "Well, to make an educated answer I'd need to sketch out a scenario. Meanwhile, my best guess is to turn individual radio buttons invalid. That helps the User know exactly where they made a mistake."

JanCBRE: I kept the props.isValid to turn a whole group invalid because it can come in handy. However, I am not overly attached and happy to get rid of it if it is deemed unnecessary.

hwigmore: i have a slight vote to remove `props.isValid` unless we think/know we need it, just to simplify things / props that are passed. but i don't feel strongly!

JanCBRE: Ok. I removed the props.isValid from RadioButtonGroup

blocks-styles/radio-buttons.styl
```diff
@@ -34,10 +35,10 @@
       display: none;
       content: "";
       position: absolute;
-      top: 4px;
-      left: 4px;
-      width: 8px;
-      height: 8px;
+      top: 3.650px;
```
emma-floored: do fractional pixels make sense? this seems a little too specific

emma-floored:
I don't think this is specific to this branch (I've seen it on some Google forms as well) but here is a weird thing that is happening:
screen shot 2019-01-22 at 1 28 07 pm

blocks-styles/radio-buttons.styl
```diff
@@ -1,9 +1,11 @@
 /* The container */
 .blx-radio {
+  background-color: $ui-01;
+  width: 220px;
```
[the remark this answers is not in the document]
JanCBRE: Good point. I will remove the set width and background color.

blocks-styles/radio-buttons.styl
JanCBRE: The Zeplin specs provide the fractional heights, but I could change it to height, width = 8px, and top, left = 3px which would look like this.
screen shot 2019-01-23 at 3 34 59 pm

JanCBRE:
@emma-floored Addressed your comments.

emma-floored:
Looks good! 👍

PR #98, opened 2019-01-21
New documentation site

emma-floored:
@hwigmore @Jinsoo94 

Reorganized the documentation content to match the new specs more closely.

This is definitely not perfect - for example, since none of the type styles have been updated yet, the type here probably also doesn't match the spec. Also, I was very lazy with the side menu and didn't do any nice animations or anything for this first pass.

Also tried to do a little more with layouts and includes to hopefully reduce some of the duplication involved with editing docs pages.

Overview of changes I made to the structure
- `_config.yml`
  - I added a Jekyll collection here - "Styles", "Components", "Patterns", and "Help". Each item in a Jekyll collection should correspond with a `_`-prefixed directory (so in this case, under `docs/` there are new directories `_styles`, `_components`, `_patterns`, and `_help`).  These directories are filled with basic md files with the necessary info to generate pages for all of the respective subsections.
- `_includes/navigation.html` is probably a good place to see how the collections are iterated over & used
- `_layouts/componentSection.html` now does more - it conditionally generates the example, edge cases (none of these yet), react preview, and code snippet sections (basically everything that was formerly in `docs/components/*.html` got moved to `_includes/components/examples/*.html` and is now just included from `componentSection.html`)

Also, as I typed this, I remembered I'm missing one SVG (the toolbox on the front page) - I will ask Runi for that and add it before merging.


emma-floored:
https://app.clubhouse.io/floored/story/13469/docs-new-documentation-site-layout

hwigmore:
yay thanks for taking this on! i took a quick glance at the site and i see Modals is now missing from Patterns. is that intentional?

emma-floored:
@hwigmore oh oops, it got removed b/c modals are not in the spec for v1... but now I remember we liked having them available to demo. I can add them back.

Jinsoo94:
wow great work! Deploy preview looks good to me. This may not be related to this PR, but one thing I found odd was the hover/selection color of toggles? Unsure if this is expected style or not.
screen shot 2019-01-22 at 10 58 44 am

emma-floored:
@Jinsoo94 I didn't update toggles at all from whatever they were previously (so whatever is there now was made before we had an explicit Toggle spec) - so, not a problem! There's a separate ticket to update toggles.

hwigmore:
👍 from me! reminder to add this to a release draft

hwigmore:
hm or do we not add doc-specific updates, only changes that are used/seen externally??

emma-floored:
@hwigmore added modals back in!

emma-floored:
https://app.clubhouse.io/floored/story/13700/new-docs-new-block-flowchart

PR #99, opened 2019-01-31
Update Blocks colors

emma-floored:
@hwigmore @JanCBRE @jforzano (maybe also @hamham91 just fyi) I don't think that all of you need to review this, but I wanted to tag people on several different teams since these are pretty major changes that will affect every product using Blocks.

This changes all the [colors](https://app.zeplin.io/project/5adf8fe14c3924b93d974cba/screen/5b3402994b04d64da0cb25e8) defined in Blocks.

A few notes:
- Previously, in variables.json, I had started to define some basic variables for different components (ex. "button-primary-default-background-color: $primary-00"). My goal with this was to start setting up something similar to the "design tokens" Liz has mentioned previously ([example1](https://design.infor.com/guidelines/identity/design-tokens), [example2](https://www.lightningdesignsystem.com/design-tokens/)). Anyway, I did check Spacer, and I don't think either Spacer or Datavis/I&L was using any of those very component-specific Stylus/Sass variables, so instead I moved the specific variables over to their respective .styl files (ex. see the top of `buttons.styl`).
- When swapping out colors, I used roughly the following conversion (maybe this will help with updates in other products 😬)
```
$brand-01 -> $primary-00
**NOTE** in the new colors, there is still a $brand-xx family of colors, but
they are suuuuper Build bright green rather than the darker green we had
before, so I moved away from using that green on any of the components!
(I think the new brand green will be sparingly used, but I wanted to highlight
this discrepancy because it's a change in value for an existing variable name)

$text-01 -> $dark-00
$text-02 -> $dark-02

$ui-00 -> $secondary-00
apparently I didn't ever use $ui-01?
$ui-02 -> $secondary-01
$ui-03/$ui-04 -> $light-01
$ui-05/$ui-06/$ui-07/$ui-08 -> $light-02

$interactive -> $primary-00
$interactive-hover -> $primary-01
$interactive-shadow -> $shadow

$selection -> $focus
$error -> $danger-00
$danger -> $danger-00
$danger-active -> $danger-02
```
- When I was going through the process of updating the colors and also pulling basic properties out into variables, I ended up updating some parts of some components to match the new specs (b/c if I was gonna update the colors, might as well try to make sure they're the right ones). I tried to make sure nothing looks crazy, but I can't guarantee that every component is right - I think we should leave that for another pass when we do the tickets that correspond to updating individual components. If you see something crazy or definitely broken here, definitely let me know, but I would say don't worry about components that don't match specs exactly for now.


blocks-styles/_all.styl
```diff
@@ -15,7 +15,6 @@
 @import "dropdowns"
 @import "popovers"
 @import "icons"
-@import "loading"
```
emma-floored: this was an accidental holdover from Datavis that I think doesn't belong in Blocks but somehow I didn't notice until now?

hwigmore:
#godisanemma

hwigmore:
(will do a quick smoke test)

hwigmore:
looks great! clicked around all the different components. only found one warning in the console, for radio, but i dont think it's pr specific

screen shot 2019-01-31 at 3 17 46 pm

hwigmore:
👍

PR #101, opened 2019-02-04
add min, max and step to textfields, remove wrapping div from dropdown

MagdalenaHenke:
No ticket for this, just fixing/adding three things that I noticed while using blocks in Spacer.


react/wrappers/closeOnClick.jsx
```diff
@@ -49,16 +49,15 @@ function closeOnClick(WrappedComponent) {
 
     render() {
       return (
-        <div ref={(node) => { this.domElement = node; }}>
```
MagdalenaHenke: This wrapping div is really difficult to work with for aligning anything vertically (at least I haven't found a good solution). I'm pretty sure we can just add the ref to the component itself, can't we?

Before: the dropdowns are all lower than the other fields, cause it's hard to vertically align them as they float to the bottom in their little wrapping div
screen shot 2019-02-04 at 3 05 56 pm

After: look at how in line they are :)
screen shot 2019-02-04 at 3 06 00 pm

react/input/TextField.jsx
```diff
@@ -59,6 +59,9 @@ const TextField = (props) => {
           onKeyDown={props.onKeyDown}
           ref={props.forwardedRef}
           autoFocus={props.autoFocus}
+          min={props.min}
+          max={props.max}
+          step={props.step}
```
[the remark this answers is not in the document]
emma-floored: I think this is fine for now but there is a ticket suggesting that we create a separate number input component with some extra nice functions for number stuff (and I suppose depending on how that one is implemented, we might still want these here anyway)

react/tabs/RadioTabs.jsx
```diff
@@ -12,7 +13,7 @@ const RadioTabs = (props) => {
           >
             <input
               type="radio"
-              name={tab}
+              name={radioTabGroupName}
```
MagdalenaHenke: If there's more than one radio button group that are identical on the page at the same time (like we have a bunch of colums where the radio tabs all have the same label and looks and everything), `name` is what distinguishes the radio tab groups from each other. When they all have the same name, the browser assumes that only one item _in all of them_ can be selected at any point in time. Gave us weird buggies.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio#Defining_a_radio_group

Giving each group an individual name solves that. Maybe using Math.random is odd for that...but it works.

MagdalenaHenke:
@emma-floored or @hwigmore - would one of you have a look? (Or anyone else you'd like to delegate this to.)

react/tabs/RadioTabs.jsx
hwigmore: why not pass in a name as `props.name`? and/or the default/fallback could be something like joining all the tab names with dashes (so like if it's Projects Clients People, we'd make the name "projects-clients-people")

hwigmore:
👍

react/tabs/RadioTabs.jsx
hwigmore: related to my first suggestion, i think radio button group does require a name?

MagdalenaHenke: For 
> and/or the default/fallback could be something like joining all the tab names with dashes (so like if it's Projects Clients People, we'd make the name "projects-clients-people")

That doesn't quite solve it if you have several Radio button groups / radio tab groups on the page with identical tab names. Like in SpacerPro, we have a bunch of rows that are all identical that use RadioTabs for the "Open Enclosed" toggle:
screen shot 2019-02-05 at 10 16 41 am
With the name just being a function of the tab names, they all get the same name - and then selecting doesn't work correctly, where you end up only being able to have _one_ option out of _all_ the rows selected (so if there's like two rows, only for example in the _first_ row and first row only will be one of the options selected, all other fields in all other rows appear as unselected).

I think I leaned toward the "random number" approach so that whoever uses these components doesn't have to know how the name needs to be unique for each group - but you're right, sounds cleaner to just require the name. I'll fix that!

hwigmore:
yeah! just make a note of it!

react/wrappers/closeOnClick.jsx
```diff
@@ -49,16 +49,15 @@ function closeOnClick(WrappedComponent) {
...
-          <WrappedComponent
-            {...this.props}
-            {...this.state}
-            isOpen={this.state.isOpen}
-            toggle={this.toggle}
-            open={this.open}
-            close={this.close}
-          />
-        </div>
+        <WrappedComponent
```
MagdalenaHenke: oooh I hadn't tested the right thing. This _does_ break the closeOnClick (e.g. Dropdowns don't close anymore by clicking outside of them). But I really wanna get rid of the wrapping div. Maybe I can figure out a way to make a ref and pass it into the component. Will tinker. Or if you have any "oh, obviously this way" ideas, let me know

hwigmore: oooh damn now i'm busted for not actaully checking this myself 

would changing it from a `div` to a `span` fix your styling/spacing issue?

MagdalenaHenke: span doesn't quite fix it...but I think React might have something for me :)

hwigmore: ❗️

react/tabs/RadioTabs.jsx
```diff
@@ -32,6 +32,7 @@ RadioTabs.propTypes = {
   tabs: PropTypes.arrayOf(PropTypes.string).isRequired,
   onChange: PropTypes.func.isRequired,
   activeTabIdx: PropTypes.number,
+  name: PropTypes.string.isRequired, // needs to be unique for each tab group
```
emma-floored: While you're at it, would you mind making this prop required on `Toggle.jsx` as well? it already has `name` as a prop, it's just not required.

MagdalenaHenke: yup, will do

MagdalenaHenke:
I have a fix for this - but somehow I'm not able to push to this branch from here (git is asking whether I have the right access rights). I noticed I also can't pull the cbrebuild master. I wonder whether it has to do with the network I'm on. I'll try again from the hostel.

MagdalenaHenke:
It worked! Pushed changes. Forwarding the ref manually to the actual component. How do you like it?

hwigmore:
i generally am confused about refs. like `createRef` versus making it inline `node => this.ref = node;` versus what is happening here https://github.com/floored/blocks/blob/master/react/buttons/LinkButton.jsx#L6

but i will read up on that and test this now. always happy to get rid of an unnecessary wrapping div.

emma-floored:
@hwigmore `React.createRef()` is the new version, `node => this.ref = node` is the old version - I thought I had replaced all the old ones but I definitely may have missed some

emma-floored:
oh also @hwigmore the one you linked (LinkButton) is a way to do refs on functional components (previously, you could only do refs with classes I think?)

emma-floored:
@MagdalenaHenke dropdowns look great now, but I think you might need to make the same change for PopoverMenus - they aren't closing on click like expected!

react/dropdowns/PopoverMenu.jsx
```diff
@@ -15,7 +15,7 @@ const PopoverMenu = (props) => {
     menuClasses += ' blx-popover-is-right';
   }
   return (
-    <div className="blx-popover-wrapper">
+    <div className="blx-popover-wrapper" ref={props.forwardRef}>
```
hwigmore: oh i think it's just a typo. forwardED ref?

emma-floored: 🎣

MagdalenaHenke: OOOpes, yup, thanks! Fixed!

MagdalenaHenke:
@hwigmore  - I was more confused about refs, too, and then mostly read this:https://reactjs.org/docs/refs-and-the-dom.html, but also I might have misunderstood some stuff. You live and you learn. :D

All that Emma said about `createRef`. Also, the callback method seems easiest if you hav access to the actual element that you want to attach the ref to - which we don't here, because that dom element is in the WrappedComponent.

`ref` isn't treated like a regular `prop`, even if you pass through `...props` from one component to its child, `ref` will get lost because it's a special prop that gets passed max one level down. `React.forwardRef` is a way to continue passing down `ref` through a component that gets it from above. I could have also (I think) passed `ref` as just `ref` into the `keyControlledMenu` and then used `React.forwardRef` in `keyControlledMenu` to be able to continue passing down `ref`. Instead I just named it something different `forwardedRef` from the start so that it gets treated like a regular prop and just passed down easily.

PR #102, opened 2019-02-05
add class for checkmark icon

hwigmore:
screen shot 2019-02-05 at 3 14 07 pm

i can't figure out how to make the checkmark bigger, no matter what css i tinker with. any ideas @emma-floored @MagdalenaHenke ? do you think the svg itself needs to be altered??


emma-floored:
@hwigmore yeah maybe we need a new check SVG... although, might not be worth it to worry about right now, given the new icon spec?

But anyway, if I alter the SVG itself (doubled the size), it's too chunky:
screen shot 2019-02-05 at 6 51 14 pm

hwigmore:
when you alter the svg did you just change width and height or also viewbox?

hwigmore:
and yeah good point re: new icons. i can close this PR. i'll do an override in i&l for the time being!

PR #103, opened 2019-02-08
split to multiple packages

hamham91:
so... big changes...

This is the high-level folder structure:
screen shot 2019-02-08 at 1 57 03 pm
where:
- `blocks-base` contains the styles, fonts, svgs, etc...
- `blocks-react` contains the react components
- `blocks-docs` contains the documentation site

Both `blocks-base` and `blocks-react` have a `gulp build` command which respectively creates the appropriate style files and transpiles the react components.

`blocks-docs` has a `gulp server` and `gulp build` command which create (and serve) the documentation site.

If you make changes to `blocks-base` or `blocks-react` and want those changes to be viewable when you're running `gulp server` in `blocks-docs` you'll need to run `gulp build` in those repositories.

For right now, we'll distribute the built styles and transpiled react components by checking them into the repository.

So you've checked out the branch and want to get started:
1) Delete `./node_modules` and other folders that aren’t present on this branch (these things were previously gitignored)
2) Run `npm install -g lerna`
3) Run `lerna bootstrap`
4) Navigate to `/packages/blocks-docs` and run `gulp server`


hamham91:
Things I want to look at in the future:

- npm preinstall commands so we don't need to commit the built files in `blocks-base` and `blocks-react`
- listing react as a peer dependency for `blocks-react` so the user of the library is required to install it

packages/blocks-base/blocks.css
```diff
@@ -2194,18 +2194,6 @@ a.blx-button {
   color: #aab7b2;
   color: var(--light-02, #aab7b2);
 }
-.blx-text-field input:-ms-input-placeholder {
```
hwigmore: hm why this one specific code change in this PR? just wanna double check that this was intentional

package.json
```diff
@@ -1,50 +1,7 @@
...
-    "gulp-postcss": "^7.0.1",
-    "gulp-rename": "^1.2.2",
-    "gulp-stylus": "^2.7.0",
-    "gulp-uglify": "^3.0.1",
-    "gulp-util": "^3.0.8",
-    "postcss-css-variables": "^0.8.1",
-    "vinyl-buffer": "^1.0.1",
-    "vinyl-source-stream": "^2.0.0",
-    "watchify": "^3.11.0"
+    "lerna": "^3.10.7"
```
hwigmore: this is only needed as a dev dependency? won't the actual blocks site need it?

(maybe related: looks like building failed)

packages/blocks-docs/_javascript/AccordionPreview.jsx
```diff
@@ -1,7 +1,7 @@
 const React = require('react');
 const ReactDOM = require('react-dom'); 
 
-const AccordionSection = require('../../react/accordion/AccordionSection.jsx');
+const AccordionSection = require('blocks-react').Accordion.AccordionSection;
```
hwigmore: ooh this is nice

packages/blocks-react/dist/accordion/AccordionSection.js
```diff
@@ -0,0 +1,90 @@
+'use strict';
```
hwigmore: i don't understand why the new structure requires checking in generated files when the old one didn't?

hwigmore:
love this restructuring! can we also update the README to describe the structure and how to run things?

packages/blocks-react/dist/accordion/AccordionSection.js
hamham91: The previous method necessitated that the application figure out how to transpile blocks. There would be some build step like "run *.jsx in `node_modules/blocks` through babel". This removes that addition to the build system so the module can just be dropped into a project. A future version of this (lightly mentioned in the PR description) is that these files could be generated on an `npm install` rather than checked into the repo.

packages/blocks-base/blocks.css
hamham91: hmm might be that I didn't regenerate this file after rebasing on Lena's changes

package.json
hamham91: yep! lerna is used to connect the interdependecies between the packages. the build is failing because it's looking for a root `/docs` folder which no longer exists and neither Emma or I have the netlify privileges to investigate further (reaching out to judy on monday!)

packages/blocks-base/blocks.css
hamham91: looks like maybe it removed a duplicate entry? see the css class below the deleted one

emma-floored:
ok deploy preview is live! all we had to do was change the build command from
```
gulp build
```
to
```
gem install bundler && bundle install --gemfile=./packages/blocks-docs/Gemfile && lerna bootstrap && ./packages/blocks-docs/node_modules/gulp/bin/gulp.js build --gulpfile ./packages/blocks-docs/gulpfile.js
```
🤢

packages/blocks-react/dist/accordion/AccordionSection.js
hwigmore: ooh nice (re: not requiring the app to transpile. i kinda forgot that we did that on the other end)

thanks for explaining!

README.md
```diff
@@ -20,11 +20,26 @@ const Checkbox = require('blocks/react-index').Input.Checkbox;
...
+gulp build
+```
+
+To build the `blocks-docs` (or documentation) package, navigate to the module in `/packages` and follow these instructions:
+
+You'll want to download [Jekyll](https://jekyllrb.com/).
+
+Use ruby v2.4.4 (you might need to install this first):
 ```
 rvm use ruby-2.4.3
```
emma-floored: 2.4.4?

Also, maybe you could add a bit here about how with this new organization, changes to one of the other packages (`blocks-base` or `blocks-react`) have to get built in their respective folders first before they can be seen in the documentation site?

PR #105, opened 2019-02-14
blocks-mithril

hamham91:
Implements an initial `blocks-mithril` package which currently only exports a PrimaryButton.

There's a "Mithril Examples" section in the blocks documentation page now.
screen shot 2019-02-14 at 5 27 43 pm

### DIFF
--- a/packages/blocks-docs/_javascript/common/getMithrilPreviewComponent.jsx
+++ b/packages/blocks-docs/_javascript/common/getMithrilPreviewComponent.jsx
@@ -1,0 +1,13 @@
+const m = require('mithril');
+
+const getPreviewComponent = (caption, component) => ([
+  <span className="doc-state-title blx-ui-small" key={caption}>
+    {caption}
+  </span>,
+
+  <div className="doc-state-content" key={`${caption}-component`}>
+    {component}
+  </div>
+]);
+
+module.exports = getPreviewComponent;
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,5 @@
 /node_modules/
 /packages/blocks-base/node_modules
 /packages/blocks-react/node_modules
+/packages/blocks-mithril/node_modules
 /packages/blocks-docs/node_modules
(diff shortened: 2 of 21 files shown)


hamham91:
Implemented the rest of the buttons!

screen shot 2019-02-14 at 6 02 40 pm

hamham91:
Note: need to add in links for buttons

hwigmore:
👍

packages/blocks-docs/_javascript/common/getMithrilPreviewComponent.jsx
```diff
@@ -0,0 +1,13 @@
+const m = require('mithril');
```
joshkarges: This file looks just like the normal `getPreviewComponent`, except one requires react and the other mithril, but neither explicitly uses them.  I'm just curious if we could use the same jsx component for both?

hamham91: So you need to include both because once JSX is compiled to javascript it substitutes in `React.createElement(..)` or `m(...)`.

hamham91:
@joshkarges see my answers below!

> 1.) What's the difference between the components in the main `blocks-mithril/` directory and the components in the `blocks-mithril/dist/` directory?

`blocks-mithril/dist` contains the compiled (i.e. babelified) versions of the components.

> 2.) If Plans engineers wanted to write components without jsx (just using the normal `m` syntax), where would we put this, and are there any extra considerations we need to make?

You would write them in `/blocks-mithril/<component grouping folder>/<component name>` mirroring the `/blocks-react` layout. You do not need to do anything explicit to opt-out of JSX, but you will want to add the components to `/blocks-mithril/index.js` and `/blocks-mithril/gulpfile.js`.

> 3.) Will the jsx components be easy to apply to the Plans app after we install the `blocks-mithril` node module or is there some babel things we need to install also?

The distributed components (located in `blocks-mithril/dist`) have been compiled so there's no babelifying needed on the application side of things.

MagdalenaHenke:
@hamham91 From what I understand, mithril components in Blocks are basically frozen. Would you be ok with us closing this PR, just to prune our open PR list?

hamham91:
👍

MagdalenaHenke:
Closing this for now because it looks like mithril-in-Blocks is a castle in the sky for future times.

joshkarges:
Time to rewrite Protofit in React!

PR #106, opened 2019-02-15
specify node and npm version

hwigmore:
@emma-floored @hamham91 

i did `nvm use` and then `npm install npm@latest -g` to get an updated version of npm that won't cause package lock issues. 

if you dont have this version locally i think you can just `npm install npm@6.8.0`

should prevent people from seeing package-lock changes when they aren't expecting them


emma-floored:
👍

PR #107, opened 2019-02-15
Change focus background for and add active state to checkboxes

samfloored:
@emma-floored Is this like good? Thank you for all the hlep!


samfloored:
https://app.clubhouse.io/floored/story/13685/update-minor-checkboxes

packages/blocks-base/styles/checkboxes.styl
```diff
@@ -129,6 +134,32 @@ $checkbox-selected-focus-background-color = $primary-00;
...
   }
 
+  &:active, &.blx-active {
+    input[type="checkbox"] ~ .blx-checkmark {
+      background-color: $checkbox-unselected-active-background-color;
+      border: $checkbox-unselected-active-border;
+      &::before {
+        display: none;
+      }
+      &::after {
```
samfloored: I think this is maybe to override the focus style, for when they're both active? Worth refactoring?

emma-floored: Nah, I wouldn't worry about refactoring for now! We have a few places where we do weird stuff like this because of different state border widths :-/

emma-floored:
Looks good to me! 👍

PR #109, opened 2019-02-15
Textfield v1 updates

samfloored:
@emma-floored 

Things here:
1. I added `header-text-fields` as a second class to apply to a text field. Happy to do this another way! It really mostly just clears out styles.
2. I moved text field helper messages to underneath the input like how errors are. Also moved the class to typography, where errors are.

screen shot 2019-02-15 at 5 39 30 pm
screen shot 2019-02-15 at 5 39 20 pm

Note: I think the colors aren't 100% final, still waiting on some design wisdom next week!


packages/blocks-base/styles/text-fields.styl
```diff
@@ -5,14 +5,17 @@ $text-field-date-icon-color = $primary-00;
...
 
 $text-field-default-border = 1px solid $light-01;
 $text-field-active-border = 1px solid $light-02;
 $text-field-hover-border = 1px solid $light-02;
 $text-field-focus-border = 1px solid $focus;
 $text-field-error-border = 1px solid $danger-00;
 
+$header-text-field-background-color = $text-field-background-color;
+$header-text-field-invis-border = 1px solid transparent;
+$header-text-field-border = 1px solid #ced4d2;
```
emma-floored: oh I see, this is the new color you mentioned? the other 2 examples didn't look right?

packages/blocks-base/styles/typography.styl
```diff
@@ -235,3 +235,11 @@ h5, .blx-h5 {
   position: absolute;
   top: 100%
 }
+
+.blx-input-helper-message {
+  blxUISmall()
+  color: $light-02;
+  position: absolute;
+  top: 100%;
```
emma-floored: just curious - what does this `top: 100%` do?

samfloored: It makes it show up underneath its parent (abs pos child of a rel pos parent, so the `100%` refers to the parent height) I think.

packages/blocks-base/styles/text-fields.styl
samfloored: Yep! Sounds like it'll probably be `light-01`.

emma-floored:
https://app.clubhouse.io/floored/story/13688/update-major-text-inputs

packages/blocks-docs/_includes/components/textfields.html
```diff
@@ -124,7 +124,7 @@
   <div class="blx-text-field">
     <div class="blx-text-field-container">
       <input type="text" value="Text input">
-      <span class="blx-text-field-helper">34/150</span>
+      <span class="blx-input-helper-message">34/150</span>
```
emma-floored: might need to rename this class in the React component as well

samfloored: Is it possible it's not in any React component? Should I add something for it?

samfloored:
@emma-floored I checked with Liz, and it sounds like we want the default (`light-01`) border. Updated, rebased, and fpushed. Any other thoughts on this?

PR #110, opened 2019-02-18
Move blocks.css to /dist

hamham91:
Moved `blocks.css` to a `/dist` folder in `blocks-base` so that paths to svgs, fonts, etc work when it is required as a node module.


hwigmore:
👍

emma-floored:
👍

PR #111, opened 2019-02-20
Button updates: adding dark and loading buttons

bleveque:
[ticket](https://app.clubhouse.io/floored/story/13682/update-minor-buttons)

Question: do blocks prs usually contain dist/ changes? this one has changes to files in dist/ that I didn't even touch, so maybe those changes were discarded in a previous pr and are only being committed now? is that okay?

Some funky business going on in this branch, so I'll leave a couple comments on the diff, too!

@emma-floored @hwigmore  would you mind taking a look?


bleveque:
https://app.clubhouse.io/floored/story/13682/update-minor-buttons

packages/blocks-base/styles/buttons.styl
```diff
@@ -104,16 +121,22 @@ $button-danger-active-text-color = $secondary-00;
   background-color: $button-primary-background-color;
   border: $button-primary-default-border;
   color: $button-primary-text-color;
-  
+
+  &.blx-disabled-rgba {
+    background-color: rgba($primary-00-hex, 0.35); // using a var() in an rgba doesn't seem to work
```
bleveque: eww. this whole thing is kind of messy...the issues are:

- I want an un-dimmed loading svg, so I don't want to use `opacity` on the button, since that appears to dim every child of the button, including the svg
- so I went with an approach of using `rgba` to give each color in the button an alpha. that's kind of a bummer because I need to use rgba on each color involved, from the background to the border to the text color
- what's a further bummer is that it appears that variables defined with `var(...)` can't be used in stylus' `rgba` function, since that function is evaluated at compile time while `var` is evaluated at runtime or something, so instead of using something nice like `rgba($button-primary-background-color, 0.35)` I'm using these out-of-left-field `-hex` vars
- another bummer is that I think this `&.blx-disabled-rgba` would ideally be applied inside the `:focus` blocks as well, to make sure the button doesn't suddenly become full opacity when it receives focus

Altogether, it's a huge bummer. Maybe an easier approach would just be to keep using `opacity` and make the svgs a little more saturated to begin with, so even the dimmed version stands out a bit. Not perfect, but maybe better? any thoughts?

packages/blocks-docs/gulpfile.js
```diff
@@ -83,6 +83,12 @@ gulp.task('bundle:vendor', () => {
     .pipe(gulp.dest(path.join(__dirname, './lib')));
 });
 
+// watch for changes to blocks-react, rerun bundle:vendor on changes
```
bleveque: an attempt at a gulp watch task for blocks-react bundle changes -- any idea why the thing in the comment below is true?

packages/blocks-react/buttons/ButtonLoading.jsx
```diff
@@ -0,0 +1,59 @@
...
+    loading: 'loading-spinner',
+    done: 'loaded-check-green'
+  },
+  primary: {
+    loading: 'loading-spinner',
+    done: 'loaded-check'
+  }
+};
+
+const ButtonLoading = React.forwardRef((props, ref) => {
```
bleveque: I'm not totally sold on this `ButtonLoading` approach. Like maybe it would be preferable to designate an existing `ButtonPrimary` or `ButtonDark` as a loading button, rather than have a separate component for loading buttons, and each `Button*` component would know which loading icon classes to apply rather than having that `iconMap` above?

packages/blocks-react/dist/dropdowns/Dropdown.js
```diff
@@ -27,7 +27,7 @@ var DropdownMenu = function DropdownMenu(props) {
 
   return React.createElement(
```
bleveque: dunno where this diff came from!

hwigmore:
> Question: do blocks prs usually contain dist/ changes? this one has changes to files in dist/ that I didn't even touch, so maybe those changes were discarded in a previous pr and are only being committed now? is that okay?

it is expected that dist/ files will change. your theory about the untouched files seems correct bc if i `gulp build` the react package on master, there are a bunch of changes. gonna open that as a separate pr (and will review this soon)

packages/blocks-react/buttons/ButtonLoading.jsx
hwigmore: i am probably missing something, but one thing that is confusing me is that we don't have special components for other common use cases of icons on buttons. like for buttons that are like "+ User" we just do a regular blocks button and add `leftIcon="plus"`.  is there a reason we can't just do that with the loading icon and the success icon?

packages/blocks-base/styles/buttons.styl
```diff
@@ -48,6 +48,22 @@ $button-danger-hover-text-color = $secondary-00;
...
+$button-dark-background-color = $dark-00;
+$button-dark-hover-background-color = $dark-01;
+$button-dark-focus-background-color = $dark-01;
+$button-dark-active-background-color = $dark-02;
+
+$button-dark-default-border = 2px solid $dark-00;
+$button-dark-hover-border = 2px solid $dark-01;
+$button-dark-focus-border = 2px solid $focus;
+$button-dark-focus-inner-border = 1px solid $secondary-00;
+$button-dark-active-border = 2px solid $dark-02;
```
hwigmore: minor, but could make a `$button-border-width` variable with 2px, so if we ever change it it only has to change in one place?

packages/blocks-react/buttons/ButtonLoading.jsx
hwigmore: another point of confusion is that this conflates loading with a success state. there are use cases where we'd want the success icon where it wasn't immediately after loading (like a click to copy to clipboard)

packages/blocks-base/styles/buttons.styl
emma-floored: hmmmmmmmmmmmm

packages/blocks-react/buttons/ButtonLoading.jsx
bleveque: sorry it's taken me a while to get back to this...good points!

for your first point, my guess is that it's a common enough use case, with enough different icons/classes necessary for different combinations of {primary,secondary,dark} and {loading,done}, that maybe it's useful to wrap it in a helper component? thoughts? that said, I definitely hear that confusion; I can double check that we really want a separate component vs just having examples of loading buttons in the blocks docs for reference

for your second point -- true! I will bring this up when asking about your first point

emma-floored:
Sorry I let this linger @bleveque !! I rebased it since a lot of changes have happened and will take a new look at it tomorrow.

bleveque:
@emma-floored oh no worries! I've been letting it linger, too -- I have an outstanding action item to check in with design about the loading button component

bleveque:
and thanks for rebasing!

packages/blocks-base/styles/buttons.styl
bleveque: good call!

bleveque:
action items from chatting with Heather:
- double-check that changing z-index doesn't affect how opacity is applied (that would be super nice...)
- next steps:
  - first update existing ButtonLoading component to use new icon convention (passing in icon component, fill being set in .blx-primary (e.g.,) styling) and see if, from there, it would be easy for people to create loading buttons without a component
  - _if_ we keep it a component, think about renaming from Loading to something like Progress (?), which could make more sense for cases like "copy to clipboard" where you want to show a done state but there's no loading involved

Probably won't get back to this imminently...

hwigmore:
great summary!!

emma-floored:
rebased on new blocks docs changes

emma-floored:
not totally sure why deploy is failing, will try to fix tomorrow

emma-floored:
Ok I think the issue is that we need to also run a `lerna bootstrap` as part of the deploy command in Netlify. But, to do that, I changed the base directory in Netlify to blocks/ (rather than blocks/packages/blocks-docs) and then it didn't have access to Gatsby. So, I might need to have Gatsby as a root level dependency? Or something else...

emma-floored:
Deploy preview fixed!

bleveque:
@emma-floored let me know if this is a blocker, and I'd be happy to revisit once build prototyping wraps up

emma-floored:
@bleveque it's not! I just wanted to update it in preparation for the Blocks mini-sprint in late June. We can either take it over then or you can fix it up - either is fine!

bleveque:
okay great, sounds good, thanks!

emma-floored:
Closing this so I can split it into 2 PRs (one for dark buttons, one for loading buttons!)

emma-floored:
ref: #158 #159

bleveque:
cool, thanks @emma-floored!

PR #113, opened 2019-02-26
Make RadioButtons valid by default

hamham91:
Make RadioButtons valid by default so they don't show up as invalid if an `isValid` flag isn't passed.

@emma-floored @hwigmore


hwigmore:
👍

emma-floored:
👍

PR #115, opened 2019-03-01
rename dropdown description to label for consistency with other input components

hwigmore:
breaking change, renaming a prop from `description` to `label` for consistency with textfield

fyi i'm leaving for a week so feel free to add to this and/or merge

@emma-floored


packages/blocks-react/dropdowns/Dropdown.jsx
```diff
@@ -27,12 +27,8 @@ const DropdownMenu = (props) => {
     <div className="blx-dropdown-wrapper" ref={props.forwardedRef}>
       <div className="blx-dropdown">
 
-        {/* DESCRIPTION */}
-        { props.description &&
-            <div>
```
hwigmore: by removing this surrounding div, the label margin-bottom actually gets applied and therefore the space between label and input is consistent with textfield

hwigmore:
@hamham91 also tagging you on this, we'll need this pulled in to I&L to get the proper spacing between dropdown label and input

packages/blocks-react/dropdowns/Dropdown.jsx
```diff
@@ -111,9 +107,9 @@ DropdownMenu.defaultProps = {
   options: [],
   text: 'Choose an option',
   icon: null,
-  description: '',
+  label: '',
   value: null,
-  onChange: () => {},
+  onSelect: () => {},
```
emma-floored: does this change fix that other bug you mentioned?

hwigmore: so i'm not 100% sure that i was right about this bug. i need to revisit this. 

this change was just changing to what we actually expect in this file (if you search the file for `onChange` you'll see it's only mentioned in propTypes, but `onSelect` is actually the prop that's used in render

related: i would love to get linter working for blocks!!

hwigmore: but i think my original bug / confusion stems from `keyControlledMenu` where there's a component function called `onSelect` that either calls `onClick` or `onChange`. i'll take a look again later, i might just be confused!

emma-floored:
👍

PR #120, opened 2019-03-14
Blocks v1 remove default margins

emma-floored:
@hwigmore @joshkarges Remove default margins (most margins were on type styles, so that's what mostly changed).

I added a `blx-with-margin` class that, when added to a component or type style, will add whichever margin is specified in the specs (not all components have a default margin specified, so `blx-with-margin` will do nothing currently for those components).

**NOTE** I also created, but did not commit, an override style sheet to help products when they upgrade Blocks to whichever release this change will appear in. Since some products have already updated Blocks to a version with margins, and accounted for those margins, copying the styles in the override sheet should hopefully keep their spacing the same so they don't have to update everything all at once.


emma-floored:
https://app.clubhouse.io/floored/story/14476/remove-default-component-margins

PR #121, opened 2019-03-15
update Toggle: add onChange, pass value instead of isChecked

hwigmore:
@emma-floored i'm not 100% sure how we want to do this. the toggle component is essentially the same as radio button group, except for styles. for radio button group we pass an index, but i thought it felt more correct to pass a `value` instead (more like a dropdown i guess, where you map and see if `option.value === props.value`)

what do you think?


emma-floored:
👍

PR #122, opened 2019-03-19
Blocks v1 tooltips

emma-floored:
@hwigmore @Jinsoo94 First pass at tooltips. There might still be some weird stuff (like they delay a big before opening?) but this gives us a starting point, at least.

Diff is big b/c I ended up tweaking the structure of the HTML code snippet sections on the doc pages (I found that the way it was previously _really_ didn't work for tooltips). Most important thing for y'all to look at probably would be Tooltip.jsx?

**Edit** also I know the icons look different - I'm gonna fix that in a follow up PR that focuses on icons


emma-floored:
https://app.clubhouse.io/floored/story/10777/new-tooltips

Jinsoo94:
I don't know if this should be the final draft or not, but I will review as if this is the final draft! Haven't looked at the code yet.

1. Left tooltip is covering 'Left' text.
Screen Shot 2019-03-21 at 4 32 20 PM

2. Dark background is too wide.
Screen Shot 2019-03-21 at 4 33 35 PM

3. React example seems to be not working properly.
Screen Shot 2019-03-21 at 4 34 35 PM

emma-floored:
@Jinsoo94 re: 3 - did you click them?

Jinsoo94:
@emma-floored 😆 ohhhhhhh I did **not** and when I do the popups show up.

packages/blocks-react/tooltips/Tooltip.jsx
```diff
@@ -0,0 +1,47 @@
...
+}
+
+Tooltip.propTypes = {
+  style: PropTypes.object,
+  className: PropTypes.string,
+  toggle: PropTypes.func.isRequired,
+  trigger: PropTypes.object,
+  lightBg: PropTypes.bool,
+  position: PropTypes.string, // 'bottom', 'top', 'right', 'left'
+  text: PropTypes.string
```
Jinsoo94: proptype is missing title

Jinsoo94: is style used anywhere?

emma-floored: oops 🎣

Jinsoo94:
I have only looked at tooltips.styl and ToolTips.jsx and they look good to me! I was wondering why you used transform: translate instead of like.. left: -50% but then I realized you want to translate by 50% of the message box, not the container.. so all good!

emma-floored:
@Jinsoo94 Fixed 1 & the missing props, leaving 2 for now b/c the background isn't part of the tooltip, it's just a way to be able to see the white tooltip.

Thank you for the review! 💐

PR #124, opened 2019-03-21
Close on click update

helenCBRE:
don't have a ticket - follow up fix to closeOnClick pr I recently made


helenCBRE:
@emma-floored mind taking a quick look? i basically just wrapped my event in an undefined check bc sometimes that event is undefined

emma-floored:
@helenCBRE what is the case where `e` is undefined?

helenCBRE:
in a dropdown for example - when i select something in the dropdown rather than click somewhere outside of the dropdown to close

helenCBRE:
oh hmm. how would we send an event when there isn't one happening from the interaction on the page?

emma-floored:
sorry, led you astray! you can fix this by passing `e` here: https://github.com/floored/blocks/blob/master/packages/blocks-react/wrappers/keyControlledMenu.jsx#L41

and also just to be safe, here as well https://github.com/floored/blocks/blob/master/packages/blocks-react/wrappers/closeOnClick.jsx#L40

helenCBRE:
@emma-floored mind taking a quick look?

emma-floored:
uh oh something strange happened

emma-floored:
with the update/merge

emma-floored:
but your changes in commit de8eba0060ab8f6ff609033cf08f33dd6a7369ba look good

helenCBRE:
woah i didn't even notice the massive file change until now rip. i rebased onto master before committing? i think i'll just close this pr and reopen it

PR #126, opened 2019-03-25
first pass at icons

emma-floored:
@hwigmore @Jinsoo94 @helenCBRE 

Did a first pass at some SVG icons. I ended up copying a lot of the svg content into React icons (mostly see the base Icon.jsx, all others use that one).

This _should_ remove all instances where we were using CSS `mask` for icon stuff (unless I missed something) - I haven't tested this yet in IE/Edge, but it should work better than our previous solution.

~One bug I know about already - the "success" alert icon is shifted funny - I need to get a new one that is the right size from Runi.~


emma-floored:
https://app.clubhouse.io/floored/story/20785/new-icons

PR #127, opened 2019-03-27
Fix icon repeating

hamham91:
Fix icon repeating when not using the react icons.


emma-floored:
@hamham91 rebuild blocks-base?

hamham91:
@emma-floored done!

emma-floored:
👍

PR #130, opened 2019-03-29
text area placeholder styling

MagdalenaHenke:
Tiny fixes for things I noticed on spacer

1. `textarea` placeholder styling should be the same as `input` placeholder styling
2. we started having an issue where selecting a nested popover option would trigger clicking the parent element, too. I feel like that's not what people want in most cases. 

@emma-floored - teeny tiny look?


MagdalenaHenke:
Oh wait, do I need to run `gulp bundle` on these? I probably do, right?

PR #131, opened 2019-04-02
export base Icon component

hwigmore:
@emma-floored i'd like to export the base icon class so that if we want app-specific icons as components, we can use the same component format as the other blocks icons


emma-floored:
👍

PR #132, opened 2019-04-16
dont use content for table cell keys, to avoid react errors for duplicate keys

hwigmore:
really minor but we have a lot of the same data (or lack thereof) in our table cells and this change is to get rid of the duplicate key errors


hamham91:
👍

hamham91:
my only thought is maybe we want to have a per table key so we can support having multiple tables on the same page?

hwigmore:
oh great idea. i can add that

hwigmore:
@hamham91 added one more commit! great idea

emma-floored:
👍

PR #133, opened 2019-04-18
Make modals optionally closable

hamham91:
This PR does two things:
1) Makes modals optionally closable
2) Makes children not a required prop

Impetus for 1) is discussed here on Slack: https://cbrebuild.slack.com/archives/C881SAF9T/p1555613958016500

Impetus for 2) is that sometimes the content for a modal is being asynchronously loaded and we'd like to be able to just use `<BaseModal>{ null }</BaseModal>` without guarding to check if it's content has loaded.


emma-floored:
👍

hwigmore:
👍

PR #134, opened 2019-04-19
Extend Dropdown trigger content

hamham91:
Proposing a modification to the Dropdown trigger content.

The proposal covers two things:
1) If your dropdown options are elements (not text), this will allow for displaying the element in the trigger
2) If your dropdown options are elements, but you want to display a string instead of your element in the trigger, this will allow you to specify that string with a `triggerContent` prop

This is the design that's prompting this proposal:
Screen Shot 2019-04-19 at 2 42 30 AM


hwigmore:
👍

emma-floored:
👍

PR #136, opened 2019-04-25
SortableList breaks if props.children is not an array 

tariqporter:
`props.children` is not guaranteed to be an array. React.Children ensures that this will be an array.
It could be a function or an object for example

https://mxstbr.blog/2017/02/react-children-deepdive/

### DIFF
--- a/packages/blocks-react/lists/SortableList.jsx
+++ b/packages/blocks-react/lists/SortableList.jsx
@@ -141,7 +141,7 @@ class SortableList extends React.Component {
         ref={this.listDOM}
       >
         {
-          this.props.children.map((child, idx) => (
+          React.Children.map(this.props.children, child => (
             <li
               key={Math.random()}
               className="blx-sortable-list-item"
@@ -170,7 +170,7 @@ SortableList.propTypes = {
 SortableList.defaultProps = {
   className: '',
   style: null,
-  onDrop: () => {}
+  onDrop: () => { }
 };
 
 module.exports = SortableList;
--- a/packages/blocks-react/lists/SortableList.jsx
+++ b/packages/blocks-react/lists/SortableList.jsx
@@ -170,7 +170,7 @@ SortableList.propTypes = {
 SortableList.defaultProps = {
   className: '',
   style: null,
-  onDrop: () => { }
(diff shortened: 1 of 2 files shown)


packages/blocks-react/lists/SortableList.jsx
```diff
@@ -170,7 +170,7 @@ SortableList.propTypes = {
 SortableList.defaultProps = {
   className: '',
   style: null,
-  onDrop: () => {}
+  onDrop: () => { }
```
hamham91: can probably drop this change

packages/blocks-react/lists/SortableList.jsx
```diff
@@ -141,7 +141,7 @@ class SortableList extends React.Component {
         ref={this.listDOM}
       >
         {
-          this.props.children.map((child, idx) => (
+          React.Children.map(this.props.children, child => (
```
hamham91: cool! I didn't know about the `React.Children` util functions, it seems like we generally want to be using `React.Children.map` everywhere we're currently using `props.children.map`, yea?

emma-floored:
nice, thanks! 👍

hwigmore:
cool! FYI when you make a change to a react component you also need to run `gulp build` from `packages/blocks-react` because we currently commit our built files! so there should be two files changed in this PR

packages/blocks-react/lists/SortableList.jsx
tariqporter: Yep!

hwigmore:
(if you dont have blocks set up locally in terms of nvm version, ruby, whatever else is in the readme, i could run the build command for you this time. let me know!)

tariqporter:
> cool! FYI when you make a change to a react component you also need to run `gulp build` from `packages/blocks-react` because we currently commit our built files! so there should be two files changed in this PR

Added the dist version!

hwigmore:
nice! 👍 

after merging, please go to https://github.com/floored/blocks/releases and create a new draft release in the same format as other releases, which lists this change. we keep a running list of the changes so that when we release we know what is going into the new version

PR #138, opened 2019-05-14
Gatsby Blocks docs

emma-floored:
@hwigmore @MagdalenaHenke @hamham91 fyi

Diff is big but it's mostly a new `package-lock.json`.

Migrate Blocks documentation from a Jekyll site to a Gatsby site and change the layout of how code examples are provided.

## Motivation
- I felt like it was a bit difficult to maintain the old Jekyll site, and it was hard to reuse code
  - I ended up sticking things in what felt like weird places that were maybe hard to discover if you were new to the site
- Separate React and HTML examples made it difficult to keep examples in sync as updates happen
- Didn't have a good idea of how to show React code examples (vs. just the interactive demos)
- Separation of HTML/React demos was confusing for more complicated components (ex. dropdowns) that require JavaScript

## Benefits of new approach
- Site can be built in React (maybe easier, more flexible?)
- Maybe our Netlify deploy script can be simplified and we can no longer commit weird generated files? (Still have to investigate this... I forgot our team name on Netlify so haven't been able to sign in)
- Single source for examples that is in React and is translated to React code and HTML code, so code examples should all be in sync and easy to update - this also means the examples are all fully interactive

## Overview of new site structure:
~For now, everything is in a new directory in `packages/` called `blocks-docs-gatsby` (hopefully in the future this would just become `blocks-docs`).~

All the site stuff is under `packages/blocks-docs/src/`, then:
- `code_examples/`: all the React code examples go here. There only needs to be 1 version of every example - this React code will be displayed and translated directly to HTML for the code snippets.
- `components/`: React component site pieces (ex. navigation, headers) - probably the most interesting thing here is `snippet.js` which handles the snippet boxes that toggle between a preview, React code, and HTML code
- `css/` - site styles
- `data/` - right now this just has the navigation info
- `js_utils/` - I just have a little request utility in here for helping me get the code snippets working
- `pages/` - React components for every page in the site, organized by section
- `svgs/` - put some SVGs in here, but a lot of them aren't actually used b/c I couldn't get inline SVGs working... I ended up translating the ones I needed into React components and just using the others via CSS `background-image`

## Testing
From blocks root directory,
```
lerna bootstrap
```
(this updates the symlinks so the new blocks docs site will update blocks-base and blocks-react as expected)
Then, run:
```
npm run dev:docs
```

~Pretty sure the PR deploy won't work. To test, you will have to install Gatsby (although anyone who did any work on the CBRE Build website should have it already?)~
```
npm install -g gatsby-cli
```
~Then change into the `packages/blocks-docs-gatsby/` site and run~
```
npm install
```
~then~
```
gatsby develop
```
~You should then be able to view the site at http://localhost:8000~

## Misc
I did check with Runi and ask about how closely she feels the "real" Blocks site needs to match the specs in Zeplin - she doesn't think we need to treat the specs as exactly what the website needs to look like. I do want design input on this at some point (probably before we would deploy it "for real"), but wanted to get some eng eyes on it first.

One big-ish change to note (that Runi seems ok with) is the removal of all the explicit hover/focus states - my justification for this is that you will see them by interacting with the components, and also I thought the original layout was a bit confusing (if a button is forced to be in the focus state, then it doesn't behave as expected if you hover on it, and that could be confusing possibly).

This might not be 100% yet but it's a good start (I think I got almost everything that was included in the old site?).

## Screenshots

Component previews:
Screen Shot 2019-05-14 at 11 25 31 AM

React code:
Screen Shot 2019-05-14 at 11 25 40 AM

HTML code:
Screen Shot 2019-05-14 at 11 25 50 AM


hwigmore:
ooh so exciting! gonna test now. 

totally agree re: not showing all the explicit states. i always found that confusing too for the same reason you said. 

question: shouldn't gatsby-cli be part of dev dependencies, rather than assuming it is installed globally?

hwigmore:
>   18:11  error  Cannot query field "siteMetadata" on type "Site"  graphql/template-strings

have you seen this?

emma-floored:
@hwigmore oops yes, I will push a fix real quick

hamham91:
cool!!

noticed a couple of things when poking around that may or may not be something:

unsupported image format?
Screen Shot 2019-05-15 at 3 23 36 PM

styling seems off on the colors page
Screen Shot 2019-05-15 at 3 24 27 PM

not relevant to this PR, but I think a little padding between the text and image might be nice!
Screen Shot 2019-05-15 at 3 26 33 PM

hwigmore:
love this!! 

a few things
1.  i saw the same error as lars
2. i do think we should put gatsby-cli in devDependencies
3. can delete all the old jekyll stuff?? 

👍

emma-floored:
@hamham91 can you clarify what about the styling is off? I feel like the number headings (00/01/02/03) are shifted over too much but otherwise it looks ok to me.

I will:
- fix the image error
- add gatsby-cli to devDependencies

I would love to delete all the Jekyll stuff and just make this `blocks-docs`... I realize now I didn't actually test building for production, so I'll do that so I know what command to give to Netlify.

emma-floored:
Re: building for production (just to make a note of this) I think I will have to comment out the prop types for the forwardedRef in Dropdowns, as Gatsby tries to render the pages server side and then crashes due to this error: https://github.com/facebook/prop-types/issues/240#issuecomment-455222878

hamham91:
@emma-floored it could just be a "me" issue, but having the text go over the different colored background columns looked like an overflow issue

emma-floored:
@hamham91 you are the 2nd person to mention that, but that's what it does in the spec as well... it makes me tempted to remove the background "stripes" even though Liz mentioned them specifically in the QA tickets

hwigmore:
not just you! i had that comment on another PR but emma pointed out it comes from design

emma-floored:
ooh I also will work on doing some kind of `blocks-react` / `blocks-base` updating from `blocks-docs` so that stuff is always up to date

packages/blocks-react/tabs/LinkSubTabItem.jsx
```diff
@@ -4,16 +4,13 @@ const PropTypes = require('prop-types');
 
 const LinkSubTabItem = (props) => {
   const classes = `blx-sub-tab ${props.isActive ? 'blx-active' : ''} ${props.isVisible ? '' : 'blx-hidden'} ${props.isDisabled ? 'blx-disabled' : ''}`;
-  const link = props.useReactLink ?
```
emma-floored: Note to self for updating release notes: removed ReactLink here because apparently it was deprecated after React v15 (https://reactjs.org/docs/two-way-binding-helpers.html, https://www.npmjs.com/package/react-linked-state-adapter). Removed in this PR b/c it was crashing the Gatsby build.

emma-floored:
Pushed some updates - you should no longer have to manually `gulp build` in `blocks-base` or `blocks-react` - they will be built for you automatically when you run `gatsby develop`. This _does_ mean that you still have to stop/restart `gatsby develop` (it doesn't watch for updates in those folders) but this is still better than the current state, so I figure it's fine for now.

Also! In order for them to build properly and be symlinked, you will have to rerun `lerna bootstrap` in the base blocks directory.

emma-floored:
also updated our old pre-commit hook since it seems useful again

packages/blocks-react/tabs/LinkSubTabItem.jsx
hamham91: It won't "break" anything, but it will force full page reloads where it wasn't before, why was gatsby crashing if it was only a deprecated feature and not fully removed?

emma-floored: Maybe I misunderstood the crashing, will try again

emma-floored: for reference, what it said was that ReactLink was undefined

emma-floored: I put it back in and it worked........ will revert this change.

I want the :shifty_eyes_skeleton: emoji but github doesn't have that

emma-floored:
https://app.clubhouse.io/blocks/story/34677/automate-building-of-jsx-files-for-react-components

emma-floored:
https://app.clubhouse.io/blocks/story/36469/tabs-on-site-fix-top-level-vertical-tabs-should-be-accordion-tabs-missing-icons

PR #146, opened 2019-06-17
Blocks v1 update dropdown styles

emma-floored:
Update dropdown and popover styles (the new design had the menu and trigger connected together with no space in between)


emma-floored:
https://app.clubhouse.io/blocks/story/34624/dropdowns-fixes
https://app.clubhouse.io/blocks/story/34526/popover-fixes

Jinsoo94:
looks good to me 👍

MagdalenaHenke:
Having a look now

MagdalenaHenke:
Ah, one thing I notice on here is that when a dropdown has an error style, we don't see its focus style anymore. Maybe change the precedence of those?

MagdalenaHenke:
Gonna mention another nitpick just cause I'm testing dropdowns here, but doesn't have to happen here: I think after selecting an item via keyboard, the visible focus should return to the dropdown trigger box ([looked here](https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-collapsible.html)).

packages/blocks-base/styles/dropdowns.styl
```diff
@@ -116,25 +120,26 @@ $dropdown-icon-spin-duration = 200ms;
   position: absolute;
   top: 100%;
   width: 100%;
-  border-radius: $base-border-radius;
+  border-radius: 0 0 $base-border-radius $base-border-radius;
   z-index: 1;
-  margin: 8px 0;
-  padding: 8px 0;
   list-style-type: none;
+  margin-top: -1px;
```
MagdalenaHenke: Side effect (or is that the intended thing?) that error-state open dropdowns look like this with the red only on three sides: Screen Shot 2019-06-18 at 12 48 46 PM

Is that what we want?

MagdalenaHenke:
Oh also nitpick but those popovers are missing a focus style.

MagdalenaHenke:
Things like closing the dropdown on `escape` - when do we worry about that?

packages/blocks-base/styles/dropdowns.styl
```diff
@@ -116,25 +120,26 @@ $dropdown-icon-spin-duration = 200ms;
...
+  overflow: hidden;
 
   background-color: $dropdown-default-background-color;
   border: $dropdown-default-border;
   box-shadow: $default-box-shadow;
 }
 
 .blx-dropdown-list {
-  max-height: 200px;
```
MagdalenaHenke: Question about this - is this a verbal thing Craig updated?

The current spec includes these sentences:
	•	A popover should almost always show all its options. If that’s really hard, the list can scroll, but it’s more important for popovers to show everything than it is for dropdowns.
	•	If the list scrolls, the list’s height should allow half of an option to show at the bottom. 

and makes me think that popover menues should usually show all options, unless there are more than like 9 options, and then we make it shorter and scroll. Since that's difficult, was the outcome of the talk with Craig to just always make it a smaller scrollable one? If yes, where do we document things like that for blocks?

MagdalenaHenke:
Well ok I think I'm done with my rant of nitpicks. Which I guess mostly are questions in the first place.

packages/blocks-base/styles/dropdowns.styl
emma-floored: ahh this is because I set the menu margin as `margin-top: -1px` so we wouldn't have a double border where the trigger meets the menu... I can rethink that strategy

packages/blocks-base/styles/dropdowns.styl
emma-floored: There was confusion with the current spec b/c of how it's hard engineering-wise to do what was written - Craig has updated the spec and I can add to this PR

emma-floored: oh but also 140 = 40px * 3.5 items fyi

emma-floored:
re: popover triggers missing a focus style - yep! this is something I brought up with Craig yesterday. I think he has updated the spec so I can add those and update this PR.

emma-floored:
https://app.clubhouse.io/blocks/story/38597/update-list-length-rules

PR #148, opened 2019-06-17
update gulp for blocks-base build

emma-floored:
@Jinsoo94 this isn't ready for review yet (still have to update `blocks-react`) but here is updating gulp for the `blocks-base` package


emma-floored:
https://app.clubhouse.io/blocks/story/38179/upgrade-gulp

emma-floored:
@Jinsoo94 @MagdalenaHenke ready now! The diff is big but it's mostly package-lock changes... One thing though is I did move all the components in `blocks-react` to a `blocks-react/components/` folder to make my gulp task a little easier (and now we don't have to remember to update the list in the gulpfile when we add a new component type!)

MagdalenaHenke:
Having a look at this now.

packages/blocks-base/gulpfile.js
```diff
@@ -1,54 +1,109 @@
...
-          .pipe(concat('variables.styl'))
-          .pipe(gulp.dest('./styles'));
-});
-
-// compile separate Blocks Stylus files into a final CSS output file
-gulp.task('build:blocksCSS', () => {
-  gulp.src('styles/_all.styl')
-    .pipe(stylus({
-      'include css': true
+function buildStylusVariableLine(key, value) {
```
Jinsoo94: ohhh so this is what you mentioned yesterday about writing your own sass/stylus parsing functions

packages/blocks-docs/src/css/_all.styl
```diff
@@ -1,8 +1,8 @@
...
+@import "~blocks-base/dist/variables"
 @import "color-cards"
 @import "home"
 @import "layout"
 @import "left-nav-panel"
 @import "page"
-@import "snippet"
\ No newline at end of file
+@import "snippet"
+@import "./syntax.css"
```
Jinsoo94: is this remnant of yesterday's deploy preview fix?

emma-floored: yeah, I don't think it actually helps but I thought maybe the ~ syntax was a little nicer than navigating to node_modules...

MagdalenaHenke:
with my global gulp conflict being dealt with, this looks and tests fine to me 🌮

MagdalenaHenke:
Oh wait should we delete variables.sass from within `/styles` then?

emma-floored:
@Jinsoo94 fyi I added an uglify and source map step to the gulp task for building the React components - they should be much smaller now!

PR #160, opened 2019-06-19
remove focus state on click

Jinsoo94:
kinda hacky solution :p would be happy to hear what y'all think! @emma-floored @MagdalenaHenke 

I didn't want to use state just for this, because
1. I didn't want to change this nice pure component to a class component
2. I didn't want re-render happening so often. (switching between true false frequently in state would trigger re-render at the same frequency)


Jinsoo94:
https://app.clubhouse.io/blocks/story/37428/ux-bug-make-tabs-focus-state-better

packages/blocks-react/components/tabs/RadioTabItem.jsx
```diff
@@ -12,8 +13,17 @@ const RadioTabItem = ({ style, className, text, children, ...other }) => {
         <input
           type="radio"
           {...other}
+          onFocus={(e) => {
+            if (mouseClicked) {
+              e.target.blur();
```
MagdalenaHenke: I am relatively strongly against this because it's easy to accidentally do something that makes this a very frustrating experience for keyboard navigation. E.g., when you're on a radio button list and instead of just navigating to the button with arrow keys and then tabbing away you also hit space "for good measure" (it's a common interaction, or used to be a common interaction that you enable/disable a button via spacebar), that triggers a focus and then this blur, and swoopedy swoop you just lost your keyboard accessibility / moved focus away from the page.

I think it's a "nifty" workaround to try and only show focus states when people interact with the page fully only with keyboard. I think it doesn't help usability (some people switch back and forth between keyboard and mouse navigation just cause of convenience), and also personally I'm not a fan of doing something iffy accessibility wise just to get rid of "ugly" focus states. I think the solution is to have more visually pleasing focus states. 

That's my two cents. I'd personally want more conversation around this before it gets merged.

emma-floored: I agree with @MagdalenaHenke on basically all of this - I think this is a wider conversation that we need to bring design in on

Jinsoo94: i hated my implementation here. I would happy to shoot it down. (I don't mind focus states, this pr was to appease those who didn't like focus states)

MagdalenaHenke:
Are you ok with closing this PR for now? I think it'll come down to 
1. more convo with the designers about whether this is really necessary
2. taking a deeeeep look into how this is "commonly" done (I tried to argue with Dan for example and he pointed out how "a lot of other sites do this", and I wonder whether anyone is doing it well consistently across all browsers)

I don't mind terribly, I just like not having PRs hang around that won't have anything happen in them soon.

Jinsoo94:
yup we can close it!

MagdalenaHenke:
Closing this PR because there still needs to be either
a) more grander discussion with Design around how we deal with those pesky keyboard focus styles, doing a bit of push back
b) more research / looking into what best practices are for what other sites do that are doing this well.

PR #163, opened 2019-06-19
use react link in dropdown

Jinsoo94:
[Clubhouse](https://app.clubhouse.io/blocks/story/34607/allow-dropdown-items-to-take-usereactlink-prop-to-avoid-refresh)

Untested because docs site yells at me for not having a router. Maybe it works on applications? Seems like LinkButton using ReactLink is also missing from docs site. Tried make one, it yells at me again.


Jinsoo94:
@emma-floored @MagdalenaHenke

emma-floored:
yeah I guess I can't test either.. but will trust Heather that we want this

👍

PR #165, opened 2019-06-20
Alert refactor: use `type` prop instead of `ErrorAlert`/`WarningAlert` etc

MagdalenaHenke:
Similar to part of what Emma just refactored with Buttons. 
First step (in this PR) is getting rid of `ErrorAlert` and `WarningAlert` and whatnot, and instead passing a `type` prop into `Alert.jsx`. 

I figured I'll separate that in review from the next step (using children for Alerts), cause it's easier to review that way.


packages/blocks-docs/src/code_examples/ButtonIconExample.js
```diff
@@ -1,7 +1,7 @@
 import React from "react"
 
 const { Button } = require('blocks-react').Buttons;
-const { UpIcon, MoreIcon, SuccessIcon, CloseIcon, AddIcon, BackIcon, NextIcon, DownIcon } = require('blocks-react').Icons;
+const { UpIcon, MoreIcon, SuccessIcon, CloseIcon } = require('blocks-react').Icons;
```
MagdalenaHenke: just noticed these were unused

packages/blocks-react/components/alerts/Alert.jsx
```diff
@@ -2,8 +2,33 @@ const React = require('react');
 const PropTypes = require('prop-types');
```
MagdalenaHenke: renamed this from AlertBanner to just Alert since it's at the front line now.

MagdalenaHenke:
@emma-floored @Jinsoo94 - first step of Alert refactor gooood to go!

Jinsoo94:
looks good to me!

emma-floored:
👍

PR #167, opened 2019-06-20
Alerts: use children instead of props and add Alerts with buttons

MagdalenaHenke:
[Clubhouse](https://app.clubhouse.io/blocks/story/38687/have-alerts-use-children-vs-props)
[Clubhouse - Alerts with buttons](https://app.clubhouse.io/blocks/story/34418/alerts-fixes)

[Zeplin](https://app.zeplin.io/project/5adf8fe14c3924b93d974cba/screen/5b43ae0ddd03f30b025bd725)

Doing two things: 
1. refactoring Alerts to use a children based structure instead of passing content props. Still figured in most cases it's convenient for a BasicAlert to be able to just pass `title` and `message`, so kept a convenience component around for that.
2. adding alerts with buttons


packages/blocks-react/components/alerts/Alert.jsx
```diff
@@ -63,17 +64,14 @@ class Alert extends React.Component {
...
-            {this.props.title}
-          </h4>
-          <p>{this.props.message}</p>
-        </div>
         { this.props.closable &&
           <button className="blx-alert-close" onClick={this.onClose}>
             <CloseIcon />
           </button>
         }
+        <div className="blx-alert-content">
```
MagdalenaHenke: moving this after the close icon for tab order (put the close icon first in tab order)

packages/blocks-base/styles/alerts.styl
```diff
@@ -1,3 +1,10 @@
+.blx-alert-wide-container {
+  width: 100%;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+}
```
MagdalenaHenke: So far, alerts just are as wide as their content needs them to be (up until a max height). The spec on Zeplin has short text, but still wide alerts - so I thought I'd add this alert wrapper of width 100% for cases like that.

packages/blocks-base/styles/modals.styl
```diff
@@ -35,11 +35,11 @@ $modal-overlay-background-color = $light-02;
...
   }
+}
 
-  .blx-modal-buttons {
-    display: flex;
-    justify-content: flex-end;
-  }
+.blx-modal-buttons {
+  display: flex;
+  justify-content: flex-end;
```
MagdalenaHenke: ok to reuse `blx-modal-buttons` in alerts? They look identical

packages/blocks-react/components/alerts/AlertBasic.jsx
```diff
@@ -0,0 +1,29 @@
...
+
+function AlertBasic(props) {
+  const { title, message, ...other } = props;
+  return (
+    <Alert {...other}>
+      <AlertBasicContent
+        title={title}
+        message={message}
+      />
+    </Alert>
```
MagdalenaHenke: in theory this is just a convenience file. Spacer could also just always do 
```jsx
<Alert> 
   <Content ...
```

but I thought it's more convenient to have the ability to just do `AlertBasic` since that's what it'll be in most cases. (But _if_ they have some custom alert content, they just use `Alert` and can pass that in themselves.) How's that?

MagdalenaHenke:
@emma-floored @Jinsoo94 - what do you think about this one?

packages/blocks-base/styles/modals.styl
emma-floored: seems fine to me

packages/blocks-react/components/alerts/AlertBasic.jsx
MagdalenaHenke: Ah per conversation right now - maybe we _don't_ want this and people really stack Alert and AlertContent themselves. Then same with ButtonAlerts - I can move the Buttons into their own AlertButtons component that people can choose to include if needed. Oooooor just have them use `Buttons` themselves?

emma-floored: I think I would vote they use buttons themselves

Jinsoo94:
is there an example for 
```
<Alert>
  <WhateverContent />
</Alert>
```
? (not using AlertBasic)

packages/blocks-react/components/alerts/AlertBasic.jsx
Jinsoo94: same! With this change to use `children`, I want individual Blocks component to be as dumb and concise as possible, much like plain HTML tags

packages/blocks-docs/src/code_examples/AlertWithButtonsExample.js
```diff
@@ -0,0 +1,70 @@
...
+
+const AlertWithButtonsExample = () => (
+  <div className="blx-alert-wide-container">
+    <Alerts.Banner type="success" closable>
+      <Alerts.Content
+        title="Success Title"
+        message="Success explanation"
+      />
+      <div className="blx-h-button-group blx-modal-buttons">
+        <Button color="secondary" onClick={() => {}}>
```
MagdalenaHenke: Ok - so got rid of `AlertsWithButtons` and instead people would have to do that like this - pop in buttons themselves (or whatever they wanna do). I feel fine with this.

I kept `Alerts.Content` around because I think that's _nicer_ than people to have to use `<h4` `<p` themselves. I think having `Alerts.Content` will help keep Alerts looking consistent. 

Like Modals, people _could_ pop anything into Alerts - there is no spec for that yet, but I think the advantage of moving to the children based approach is that we're giving eng flexibility to cases that we _aren't_ thinking of yet. But I've also looked at Modals in Spacer, and at one point in time they all had different headings, because one could just pop in any h tag one wanted. My hope is that having an `AlertsBasicContent` component will cover most cases and keep styling consistent. 

@Jinsoo94 @emma-floored - what do you think?

emma-floored:
👍

PR #172, opened 2019-06-20
Blocks color list

emma-floored:
@Jinsoo94 @MagdalenaHenke thoughts?


emma-floored:
https://app.clubhouse.io/blocks/story/38734/export-some-color-list

Jinsoo94:
is this only used for Icon (so svg) fill colors?

Jinsoo94:
I like it! One thought would be, since these are constants, I would prefer them to be in uppercase?
ex)
```
COLORS.DARK_00 = 'var(--dark-00)';
COLORS.DARK_01 = 'var(--dark-01)';
```

emma-floored:
Oh good point @Jinsoo94, I will change.

One thing I did sort of wonder was whether this belongs in `blocks-base` instead of `blocks-react`? But I wasn't sure...

Jinsoo94:
hmmmmmm my gut says `blocks-base` since `colors.js` has no react in it.. but unsure

MagdalenaHenke:
Ooooh cute. 👏

I don't have a strong preference for whether it's in base or in react. 
I think since it's all css and just providing a nice way to access css variables, it makes sense in base.
But since it'll only be used together with react things, it also makes sense to me there.

emma-floored:
- moved colors list to blocks-base
- wrote a function that will check for IE and then return color values from variables.json based on the result (`$color-name-hex` if IE, the CSS variable if not)

MagdalenaHenke:
Before this gets merged (mostly a reminder to myself), it'll need an `npm run build:react`.  (I think it's kind of nice to not clutter the PR with those changes and just run that once at the end, but I also know I'd forget to make sure that happens).

emma-floored:
@MagdalenaHenke rebuilt the React package, rebased on master, and smooshed!

packages/blocks-base/gulpfile.js
```diff
@@ -28,7 +28,7 @@ function buildSassVariableLine(key, value) {
 }
 
 function buildStyleVariablesFromJson(cb) {
-  const variableJson = require('./styles/variables.json');
```
MagdalenaHenke: I guess I'm just leaving a comment to show that I thought about things and justify why this took time: I made extra sure that moving around `variables.json` in blocks-base doesn't affect consumers like Spacer. I had a vague memory that we import variables directly from blocks so we can access them, which we do, but we get it from /dist (`@import '../../../node_modules/blocks-base/dist/variables.sass'`) so all is well

packages/blocks-base/colors/Colors.js
```diff
@@ -0,0 +1,51 @@
...
+  if (isIE) {
+    return hexValue;
+  }
+  const pieces = name.split('-');
+  let color = variablesJson[pieces[0]];
+  if (pieces.length > 1) color = color[pieces[1]];
+  return color.replace(new RegExp(/\$(.*)hex/, 'gm'), hexValue);
+}
+
+const Colors = {};
```
MagdalenaHenke: Well, only thing is that maybe a comment somewhere in this file would be good for why we do this. Cause future us will forget / new folks will be confused.

emma-floored: good call

PR #174, opened 2019-06-26
rest operator for buttons!

emma-floored:
@Jinsoo94 @MagdalenaHenke mind scoping this out real quick?


packages/blocks-react/components/buttons/Button.jsx
```diff
@@ -28,7 +27,6 @@ const BasicButton = React.forwardRef((props, ref) => {
 BasicButton.propTypes = {
   className: PropTypes.string,
   text: PropTypes.string,
-  isDisabled: PropTypes.bool,
```
Jinsoo94: shouldn't `disable` also have prop validation since it is consumed in this component?

Jinsoo94:
looks good to me!

PR #175, opened 2019-06-26
update react-dom as well so we can use hooks

emma-floored:
@Jinsoo94 @MagdalenaHenke oops I think I should have done this in #170


Jinsoo94:
lol 👍

PR #179, opened 2019-08-16
Hooks for accordion

Jinsoo94:
yeaaaa hooks


Jinsoo94:
@emma-floored @MagdalenaHenke feedback welcome!

README.md
```diff
@@ -70,3 +70,12 @@ npm run serve:docs
 ```
 from the root directory.
 
+### Configuration to use React Hooks
+
+As of React 16.8.x, React community introduced hooks which enable state management in functional components. Functional component has many benefits compared to its class counterpart; if you are curious, you can read about the motivation behind hooks [here](https://reactjs.org/docs/hooks-intro.html#motivation). To keep Blocks up to date with the industry trend, we will gradually replace class components in Blocks with functional components using hooks where applicable.
+
+If this is your first time cloning or pulling after the introduction of hooks, when you `npm run dev:docs` and try to open dev page, it will complain about duplicate instance of React. This is due to the fact that `blocks-react` and `blocks-docs` each having its own instance of React; when `blocks-docs` references `blocks-react`, it results in two instances of React in a single project. To resolve this, you can `npm i npm-link-shared -g`, then in root `blocks/` directory run the following command.
```
emma-floored: Does `npm-link-shared` have to be installed globally or should it be a devDependency?

README.md
```diff
@@ -70,3 +70,12 @@ npm run serve:docs
...
+```
+npm-link-shared ./packages/blocks-docs/node_modules/ ./packages/blocks-react/node_modules/ . react
```
emma-floored: Could create a script for this in `package.json` so it's easier to do? Then we could just run `npm run <script name>` to do this line and not have to always copy/paste it from the README.

README.md
emma-floored: ah I see, never mind!

emma-floored: maybe add a bit to the README about having `yarn` installed, as well? I didn't have it and got some errors!

README.md
Jinsoo94: done!

Jinsoo94:
@emma-floored updated readme and added script to package.json

README.md
MagdalenaHenke: lets update the text in the README then, too, to just run that command

packages/blocks-react/components/accordion/AccordionSection.jsx
```diff
@@ -14,39 +14,31 @@ function onToggle(e) {
...
-      this.content.style.maxHeight = `${this.content.scrollHeight}px`;
-    }
-  }
-
-  render() {
-    return (
+const AccordionSection = (props) => {
+  const { open } = props;
+  const contentRef = React.useRef();
+  React.useEffect(() => {
```
MagdalenaHenke: I haven't tested this yet, but I wonder whether we wanna use a `useLayoutEffect` here, since it's literally dom manipulation of things that might be visible, and I thiiiink otherwise we could end up with some flashing?

MagdalenaHenke:
Two general questions:

a) So I see this be used in Datavis, but I remember talking to you @emma-floored about why this wasn't like "official" yet. Is that something I can figure out? Like I can check with Runi and have this one approved + put it into documentation (or iterate on it depending on what needs fixing). But maybe you remember more.

b) @Jinsoo94 just curious how you went about testing this? I wanna find a setup so I can test PRs like this in Spacer / Lars mentioned a setup of like npm linking to my local blocks package from Spacer. Is that what you're doing? Or how are you testing? Just curious so I know what works and what doesn't and can benefit from what you've tried already :D

README.md
Jinsoo94: oki

packages/blocks-react/components/accordion/AccordionSection.jsx
Jinsoo94: i think you are right!

Jinsoo94:
@MagdalenaHenke I just check the docs page

MagdalenaHenke:
> @MagdalenaHenke I just check the docs page

@Jinsoo94 - ...like you just run the docs page? Accordions aren't in the docs yet

Jinsoo94:
Screen Shot 2019-10-07 at 11 06 44 AM
These are accordions.

MagdalenaHenke:
WHOOOOOOA. Amazing. Duh. Thank you.

README.md
MagdalenaHenke: I actually now have the same question as Emma - why do it globally and not as a devDependency?

MagdalenaHenke:
@sageknives @noreenwu I'm reviving this old PR to bring Hooks into Blocks - so we could hooksify the HOCs! This PR includes some instructions in the README and then a proof of concept that it works by hooksifying Accordions.

How do you feel about this?

PR #181, opened 2019-09-04
Switched defaultChecked to checked

graydonpleasants:
basically the component was unable to be controlled externally, this fixes that keeping the same api, so don't see any breaking changes


MagdalenaHenke:
Fine! Two comments
1. I think if we update that in Toggle.jsx, it makes sense, for consistency, to also update `defaultChecked` to `checked` in RadioButton.jsx, for the same reason.
2. I don't think the changes to `/dist/` on this PR picked up the right changes. Just requires rerunning `npm run build:react`.

Otherwise this is good to go!

graydonpleasants:
@MagdalenaHenke updated my branch with the requested changes

MagdalenaHenke:
Looks good! 🌮

PR #182, opened 2019-09-05
Accessible Toggle Component

graydonpleasants:
-Added aria-label and role=radiogroup

We can also just have this use the name prop, but that would not normally be screen-reader friendly


MagdalenaHenke:
Looks good to me! In general, in support of also updating the examples (like in this case `/blocks-docs/src/code_examples/ToggleBasicExample.js`) appropriately when changes to components happen, but fine in this case.

PR #184, opened 2019-09-16
add skip link to blocks docs to allow easier testing by keyboard

MagdalenaHenke:
While trying to test for keyboard accessibility, I found it cumbersome to have to tab through the entire nav section each time before reaching a component. Threw in a [skip link](https://dequeuniversity.com/tips/add-skip-navigation-link) on there - it's only visible as the first item that's being tabbed to if you navigate by keyboard, and only shows up while it has focus. It lets you skip right to the main content.

When it has focus, it looks like this:

Screen Shot 2019-09-16 at 1 55 20 PM


MagdalenaHenke:
@emma-floored or @Jinsoo94 - I did this to make my own (and everybody's) life easier when testing things for keyboard accessibility, not because of a designer. If you think the code is fine I can run it by a designer, though. If you think the code is fine! :)

For testing: go to the blocks docs, and _taaaaab_. The Skip Link will show up. If you hit enter, your focus jumps past the navigation and you can tab right into the React examples. If you keep tabbing, however, the skip link disappears again.

emma-floored:
Cute! I also found tabbing through the nav stuff tedious. Curious whether you've read anything about strategies for this - is this a common solution or just one you came up with? (I am fine with this solution, btw, just curious!)

One thing to double check out is other browsers... This worked great in Chrome, but did not work at all in Firefox, and maybe I'm dumb but just now I couldn't figure out how to do any tabbing at all in Safari?? Although, tabbing in the existing nav bar is already weird in Firefox for reasons that I don't totally understand (you can only tab to the top level items, but none of the subtabs?)

MagdalenaHenke:
1) Skip links are a common technique that I didn't make up! cbrebuild.com also has one due to some late a11y efforts there. Talked about here for example (that page also implements its own skip link): https://dequeuniversity.com/tips/add-skip-navigation-link
2) I'll have a look into browser behavior and whether that's worth fixing here. Thanks! (...kicks into the greater question of _we should probably be testing cross browser more_)

MagdalenaHenke:
Hm - my Firefox works perfectly fine for both the skip link and tabbing into the subtabs. Could you double check Firefox? I'm on firefox version 69.0.

On _Safari_ however, both is broken for me (having a little gander at why that might be now).

emma-floored:
@MagdalenaHenke hmmmmmmmmmmmmmmmmm at first I thought it must be the versioning thing (I was on a version of Firefox 68) but after updating, I still don't see the button. Either way, since you _do_ see it, probably not worth investigating in this PR?

MagdalenaHenke:
Ha - Safari sub tabs mystery is solved: _feature_, not a bug!
https://stackoverflow.com/a/2292336

> On the Mac, tab selects objects other than links, Option+Tab select all objects.

That way you can get to the subtabs and also to the skip link. It doesn't interact how one would expect (still working on that), but at least part of the mystery solved!

MagdalenaHenke:
Added a `tabIndex="-1"` on the skip link target for Safari, and now it works (if you tab to it with Option+Tab, and hit enter on the link). TabIndex of -1 means that it's not an element that's in the "natural" tab order, but can be given focus programmatically. From what I know, this is a fine use case for it.

MagdalenaHenke:
I'm fine with the Firefox situation as it is...mostly because I can't fix it without being able to reproduce 😶

MagdalenaHenke:
Runi is a-ok with this.

emma-floored:
👍

PR #187, opened 2019-09-19
isDisabled and isSelected -> disabled and selected

MagdalenaHenke:
[Clubhouse](https://app.clubhouse.io/blocks/story/41077/dropdowns-should-use-disabled-instead-of-isdisabled-for-consistency-with-other-components)

Was this just an oversight, or did we have a good reason to not do `isSelected` / `isDisabled` for Dropdowns? My hunch is - oversight.


MagdalenaHenke:
@Jinsoo94 @hwigmore - would you have a look at this one? Just an oversight or was there a reason for not changing the property names here?

packages/blocks-react/components/dropdowns/ComboBox.jsx
```diff
@@ -23,8 +23,8 @@ const ComboBox = (props) => {
   return (
     <select
       style={props.style}
-      className={`${props.isDisabled ? 'blx-disabled' : ''} ${props.className}`}
-      disabled={props.isDisabled}
+      className={`${props.disabled ? 'blx-disabled' : ''} ${props.className}`}
+      disabled={props.disabled}
       value={props.value}
       onChange={props.onChange}
```
hwigmore: can wee replace lines 27-29 (maybe more, i'm just looking at the github diff) with `{...other}` like we do here https://github.com/floored/blocks/blob/master/packages/blocks-react/components/input/TextField.jsx#L64

that way, instead of needing to update blocks whenever we add a prop/attribute, we can already use any built-in html attributes?

packages/blocks-react/components/dropdowns/Dropdown.jsx
```diff
@@ -34,15 +34,15 @@ const DropdownMenu = (props) => {
...
-          </label> 
+          </label>
         }
 
         {/* TRIGGER */}
         <button
           className={triggerClassNames}
-          disabled={props.isDisabled}
+          disabled={props.disabled}
           onClick={props.toggle}
```
hwigmore: i'd like to make the same suggestion here, but i guess our prop names dont match up (onClick is passed as toggle).

Jinsoo94:
looking now 👀

Jinsoo94:
looks good to me. I wonder why we left it out last time..

packages/blocks-react/components/dropdowns/ComboBox.jsx
MagdalenaHenke: Done! Although Combobox is entirely unused right now. Opened a PR to delete it (for now).

packages/blocks-react/components/dropdowns/Dropdown.jsx
MagdalenaHenke: Iiiiiii tried to do that. Cause in general I'm with you. 

But then I encountered an iffy thing: so here we have the trigger button (line 43) and the DropdownItem (line 64). And they get _some_ of the same props passed (like onKeyDown) and stuff. But some of the props only get passed to one of them (like `disabled` only goes to the former and `onSelect` only goes to the latter) - and if I used `...other` on either one of them, we might end up accidentally passing stuff to that one that was intended for the other one. 

So I'm with you, but I don't think it works great in this particular situation. Is that ok?

MagdalenaHenke: @hwigmore - ☝️ nudge

packages/blocks-react/components/dropdowns/ComboBox.jsx
MagdalenaHenke: Combobox is deleted now.

packages/blocks-react/components/dropdowns/Dropdown.jsx
MagdalenaHenke: @hwigmore I might just go rogue and merge this and hope my reasoning is solid enough to have convinced you.

PR #188, opened 2019-10-03
delete ComboBox.jsx, since it's unused and has no immediate plans to be used

MagdalenaHenke:
@emma-floored  - when we talked in person, you mentioned maybe we should delete `ComboBox.jsx`?

I like keeping only the things that are ready to be used / are maintained around. (I noticed it because in another PR I got some comment on ComboBox stuff and was like...welll but this is a fake component right now.)

Can also keep it if you have plans for it coming up!


emma-floored:
👍 I'm fine with deleting!

PR #189, opened 2019-10-10
[toggle fix]

graydonpleasants:
changed the className prop type to match other components, preventing an in browser error


MagdalenaHenke:
Looks good, thanks! 🌮

PR #191, opened 2019-10-14
updated docs page and action modal component

Jinsoo94:
This PR updates ActionModal to match the [spec](https://docs.google.com/spreadsheets/d/1sHDw_tVnLdjz5FOcLluC0K7cQhIIarKYcSTyNLaTZ-s/edit#gid=0).

BaseModal still works the same. It is the bare-bone modal component upon which you can build whatever modal you need. With intention, ActionModal is made more rigid. It is still flexible enough to receive a content as `props.children` in which you can put anything, but it will always have title, icon next to title that indicates the type of the modal, colored top border, and action buttons at the bottom.

Master:
Screen Shot 2019-10-14 at 7 19 18 PM
Screen Shot 2019-10-14 at 7 19 41 PM

Branch:
Screen Shot 2019-10-14 at 7 19 06 PM
Screen Shot 2019-10-14 at 7 20 04 PM
Screen Shot 2019-10-14 at 7 20 48 PM


Jinsoo94:
@MagdalenaHenke @tariqporter review pls!

Jinsoo94:
https://app.clubhouse.io/blocks/story/43065/alerts-confirmation-modals

packages/blocks-react/components/modals/ActionModal.jsx
```diff
@@ -4,15 +4,39 @@ const PropTypes = require('prop-types');
...
+    icon: ErrorIcon
+  }
+};
 
 const ActionModal = (props) => {
   if (props.isHidden) return null;
 
   let confirmButton = null;
   if (props.onConfirm && props.confirmText) {
+    const Button = props.type === 'warning' ? ButtonDanger : ButtonPrimary;
```
tariqporter: Trying to understand this. Why is `props.type` of ` warning` mapping to `danger`? Warning is usually yellow and danger (or error) is red isn't it?

Jinsoo94: haha i dont understand either but thats what the design team wants so 🤷‍♂

Jinsoo94: if you look at the spec doc I linked in the description, warning has danger(red) button

tariqporter: Oh, I see. That's definitely weird to have a warning modal with danger content, but yeh it's to spec!

tariqporter: - Especially when we have a warning button

Jinsoo94: my guess is, they want warning modal to show up in response to delete action?

Jinsoo94: since deleting is not error behavioir

packages/blocks-react/components/modals/BaseModal.jsx
```diff
@@ -7,28 +7,22 @@ const BaseModal = (props) => {
...
-      style={props.style}
-      className={props.className}
-    >
+    <React.Fragment>
       <div
         className="blx-modal-overlay"
         onClick={props.onClose}
       />
-      <div className="blx-modal">
+      <div className={`blx-modal ${props.className}`} style={props.style}>
```
tariqporter: Can use `classnames` here

https://github.com/floored/blocks/blob/master/packages/blocks-react/components/tooltips/Tooltip.jsx#L3

packages/blocks-react/components/modals/ActionModal.jsx
```diff
@@ -46,19 +70,23 @@ const ActionModal = (props) => {
     );
   }
 
+  const Icon = TYPE_MAP[props.type].icon;
+
   return (
     <BaseModal
       style={props.style}
-      className={props.className}
+      className={`${props.className} ${TYPE_MAP[props.type].className} blx-action-modal`}
```
tariqporter: and here

packages/blocks-react/components/modals/ActionModal.jsx
tariqporter: It meets spec so all good for now, but maybe something to bring up later. I know Jake feels strongly about this type of stuff. Danger should be reserved for errors that cannot be fixed

packages/blocks-react/components/modals/ActionModal.jsx
```diff
@@ -68,27 +96,27 @@ ActionModal.propTypes = {
...
-  message: PropTypes.string,
-  actionInfo: PropTypes.string,
   confirmText: PropTypes.string,
-  cancelText: PropTypes.string,
   onConfirm: PropTypes.func,
+  cancelText: PropTypes.string,
   onCancel: PropTypes.func,
-  onClose: PropTypes.func
+  onClose: PropTypes.func,
+  type: PropTypes.string,
```
tariqporter: can use

```js
type: PropTypes.oneOf(['success', 'information', 'warning', 'error'])
```

Was the property name already decided on outside this PR? I prefer `variant` over `type` as `type` already has meaning in many js contexts - `typeof` and with `Flow`s `type`, but happy to leave if you feel differently, or it was already decided!

Jinsoo94: oh ok i will switch to `variant`!

Jinsoo94:
@tariqporter updated!

tariqporter:
It all looks really good. The only funny thing is that the warning modal has an error button, and the error modal has a success button from the specs 😆 👍

packages/blocks-base/styles/modals.styl
```diff
@@ -59,13 +69,54 @@ $modal-overlay-background-color = $light-02;
...
   overflow: hidden;
+}
+
+.blx-modal-title {
+  display: flex;
+  align-items: center;
+  margin-bottom: 16px;
+}
 
-  h4 + p {
```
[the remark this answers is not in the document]
tariqporter: I also like the idea of `BaseModal` not determining content

packages/blocks-docs/src/code_examples/ModalActionExample.js
```diff
@@ -22,16 +30,34 @@ class ModalActionExample extends React.Component {
         <ActionModal
           isHidden={!this.state.modalVisible}
```
tariqporter: I guess this was already decided before this PR. Is this a large change to instead of `isHidden` make it so we use `open`. I think it's much easier to see what `open=false` does compared to `isHidden=false`

Jinsoo94: oh good point. I agree that `isHidden={!this.state.modalVisible}` is harder to read than `open={this.state.modalVisible}`

tariqporter: Reminds me of a long-running Issue for lodash I've been having 😆 
https://github.com/lodash/lodash/issues/137#issuecomment-360406431

packages/blocks-react/components/modals/BaseModal.jsx
```diff
@@ -1,34 +1,28 @@
...
+    <React.Fragment>
       <div
         className="blx-modal-overlay"
         onClick={props.onClose}
       />
-      <div className="blx-modal">
+      <div className={classnames('blx-modal', props.className)} style={props.style}>
         {
           props.isClosable && (
             <button className="blx-modal-close" aria-label="close" onClick={props.onClose}>
```
MagdalenaHenke: [This PR](https://github.com/floored/blocks/pull/190/files) suggests allowing to pass in extra classes for the close button and the overlay. That'll be conflicts with this one, so pointing it out - wanna hear your thoughts on that PR.

packages/blocks-react/components/modals/ActionModal.jsx
```diff
@@ -46,19 +71,23 @@ const ActionModal = (props) => {
...
     >
-      <div>
-        <h4>{props.title}</h4>
-        <p>{props.message}</p>
-        {actionInfo}
-        {buttonBar}
+      <div className="blx-modal-content">
+        <div className="blx-modal-title">
+          <Icon className="blx-modal-icon" />
+          <h4>{props.title}</h4>
```
MagdalenaHenke: a11y wise: I think this should be an h1 (can style with blx-h4 though).

MagdalenaHenke: There are more things a11y wise that are broken on modals right now but that doesn't have to be fixed on this PR.

packages/blocks-base/styles/modals.styl
```diff
@@ -59,13 +69,54 @@ $modal-overlay-background-color = $light-02;
...
-    margin-top: 18px;
+.blx-modal-icon {
+  border-radius: 50%;
+  width: 30px;
```
MagdalenaHenke: yeah this too...I think accessibility wise, we want to do fewer measurements in px and more in rem, for text magnification - but again that's not working on all of blocks right now either way so I'm mentioning it but it's ok if you don't want to fix it here.

packages/blocks-docs/src/code_examples/ModalActionExample.js
MagdalenaHenke: Yo I'm happy to change this everywhere, although maybe on a separate PR that's easy to document. I also don't love the "negativeByDefinition" bools.

packages/blocks-react/components/modals/BaseModal.jsx
Jinsoo94: i think it is for the right direction

packages/blocks-docs/src/code_examples/ModalActionExample.js
```diff
@@ -22,16 +30,34 @@ class ModalActionExample extends React.Component {
...
           cancelText="Cancel"
           onCancel={() => this.closeModal()}
+          variant={this.state.modalType}
+        >
+          <p>Are you sure you want to do this?</p>
+        </ActionModal>
+        <ButtonPrimary
+          text="Open Success Modal"
+          onClick={() => this.setState({ modalVisible: true, modalType: MODAL_TYPES.SUCCESS })}
+          style={{ marginBottom: 10, background: '#39cf94', width: '100%', border: 'none' }}
```
MagdalenaHenke: not a fan of this inline style, beeecause
- loses all hover and intentional focus styles, so that's not a practice I want to encourage by having it in the blocks docs
- Two of these are just a ButtonPrimary and a ButtonError so I think we should use those. The other ones...well I guess, if we don't have blocks buttons that have those colors...maybe they shouldn't be designed? Or are we adding purple blocks buttons?

Jinsoo94: I just wanted to match the color of the modals. Happy to switch them all back to ButtonPrimary.

Jinsoo94:
@MagdalenaHenke would you be ok if we handle the a11y related things in a different PR? I want this in soon because this is blocking a lot of tickets in Site IQ 😬

packages/blocks-docs/src/code_examples/ModalActionExample.js
MagdalenaHenke: Would love that.

MagdalenaHenke:
Yup, a11y stuff some other time cool with me.

packages/blocks-docs/src/code_examples/ModalActionExample.js
Jinsoo94: mind if i keep marginBottom and width? without it the docs page looks like this:
Screen Shot 2019-10-17 at 2 18 19 PM

MagdalenaHenke: oh and happy if you wanna use ButtonError or something totally down for it.

But I know eventually we should update the docs pages (whenever the design for those is "ready") so I'm cool with whatever here.

Jinsoo94:
@MagdalenaHenke updated!

MagdalenaHenke:
I'm sorry just one more nitpick. What if in `ModalExample.js` we wrap the title in blx-modal-title?
```jsx
          <div className="blx-modal-title">
            <h4>Base Modal</h4>
          </div>
```

That way at least the documentation makes people aware they can use that class for styling "out of the box". 

And other than that, 🌮

PR #193, opened 2019-11-21
babel Colors.js in blocks base, so other products using uglify don't encounter es6 syntax

MagdalenaHenke:
I admit I got a bit sidetracked from what I was _supposed_ to be working on, but this needed fixing so oh well.

Context:
In SpacerPro, we kept re-babeling the blocks packages in our own build step. I noticed that when I switched Spacer to the private repo. We'd get an error in the build step for review apps if we stopped doing it (see [this PR](https://github.com/floored/workplace/pull/1223)). Uglify would complain.

I realized it's because we used some es6 syntax in `Colors.js` in `blocks-base`, a file that we forgot to babel. Babeling it.

This PR is just making sure we also babel that file.


MagdalenaHenke:
@hamham91 - you helped me on this! Here's the fix - it's not in `block-react`, but in `blocks-base`. 
Also maybe @tariqporter.

hamham91:
👍

packages/blocks-base/package.json
```diff
@@ -18,11 +18,16 @@
   "homepage": "https://github.com/floored/blocks#readme",
   "devDependencies": {
     "autoprefixer": "^8.3.0",
+    "babel-core": "^6.26.3",
+    "babel-preset-es2015": "^6.24.1",
```
tariqporter: Just one thing. Should we be using `env` here, or do we need `es2015` for a compatibility reason?

MagdalenaHenke: I copied from what's in the `blocks-react` package.json honestly without thinking too deeply about it. If I understand it right, `env` is what's recommended now, so we should do that, but it's all fine and working with 2015. So I'm cool with us updating that (as long as we're consistent in base and react) - not on this PR though because I got too excited after Lars first 👍 and already merged :/ sorry

tariqporter:
Hehe no problem, yeh we can update to `env` at some point 👍

PR #202, opened 2020-01-20
Typescript

sageknives:
Updated blocks to use typescript.
1. Import/ require in react-docs are backwards compatible and work with es6 import destructuring
IE Examples are all valid
import { ErrorAlert } from 'blocks-react';
const {ErrorAlert} = require('blocks-react'); 
const {ErrorAlert} = require('blocks-react').Alerts;

2. type was used over interface since you can declare it below the component with propTypes and it still works on hover.
3. class based components do not show props on hover.
4. all components should have code completion.
5. there are extra files because GitHub makes the decision when you rename a file to either consider it new if a certain percentage of the file is changed. So smaller files are harder to compare since it decided I deleted a file and created a new one even though its the same file. (option: rename the file back to jsx, we can all review and comment on the code, then change back to check if it works?)
6. gulp-typescript isn't converting the file the same as before so there are a lot of file changes at the end. Possible check the to see if we can get it to compile the same? I'm not sure what is different, it looks like its not minifying or adding a source map to it?


packages/blocks-react/components/alerts/AlertBanner.tsx
```diff
@@ -1,11 +1,14 @@
...
-const CloseIcon = require('../icons/CloseIcon.jsx');
+import CloseIcon from '../icons/CloseIcon';
 
-class AlertBanner extends React.Component {
-  constructor(props) {
+/** AlertBanner*/
+class AlertBanner extends React.Component<AlertBannerProps, AlertBannerState> {
+  declare static propTypes;
+  declare static defaultProps;
+  constructor(props: AlertBannerProps) {
```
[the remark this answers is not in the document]
bleveque: also -- is everything in blocks now effectively typescripted?

packages/blocks-react/components/alerts/AlertBanner.tsx
```diff
@@ -73,4 +76,17 @@ AlertBanner.defaultProps = {
   onClose: null
 };
 
-module.exports = AlertBanner;
+type AlertBannerProps = {
```
bleveque: worth creating a followup ticket for a build step to create proptypes from ts definitions?

packages/blocks-react/components/alerts/ErrorAlert.tsx
```diff
@@ -0,0 +1,37 @@
...
+  closable: PropTypes.bool
+};
+
+ErrorAlert.defaultProps = {
+  className: '',
+  title: 'Error',
+  closable: false
+};
+
+type ErrorAlertType = React.FC<{
```
bleveque: what are your thoughts about making the types we define for functional and class components consistent? (i.e., it looks like we're defining a type for the component itself here, while we're defining a type for the class component's props above)

packages/blocks-react/components/alerts/WarningAlert.tsx
```diff
@@ -0,0 +1,37 @@
+import * as React from 'react';
```
bleveque: hm it _is_ a bit tricky to see what's changed and what should be reviewed (I suspect it's import/export, adding the type to the component signature, and adding the type definition below?) -- I'd be in favor of the option you presented in the pr description if it's not too much work for you!

bleveque:
https://app.clubhouse.io/blocks/story/47045/typescript

packages/blocks-react/components/accordion/AccordionSection.tsx
```diff
@@ -1,7 +1,7 @@
-const React = require('react');
```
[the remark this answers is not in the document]
sageknives: yes, typescript doesn't know require.

packages/blocks-react/components/alerts/AlertBanner.tsx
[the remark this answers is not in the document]
sageknives: yes, we could add more of a description to components and comments on what each prop means. We could have a follow up ticket that includes intrinsic types from whatever other is being passed to has code completion.

packages/blocks-react/components/alerts/AlertBanner.tsx
sageknives: I'm not sure. Sometimes they aren't the same because you end up writing double propTypes because its leveraging a base component. But if typescript wrote it for us, then I don't see why not? I just wouldn't do it the other way where propTypes define typescript

packages/blocks-react/components/alerts/ErrorAlert.tsx
sageknives: I thought about this. Technically this functional component is of "type" React.FunctionalComponent<props>
where above is React.Component<props,state,context>. So we are actually on defining props and state in the class because we are extending "React.Component" which has to be typed regardless of typescript. If I'm not understanding can you give me an example?

packages/blocks-react/components/alerts/WarningAlert.tsx
sageknives: done

packages/blocks-react/components/alerts/AlertBanner.tsx
sageknives: yes that would be a good idea.

packages/blocks-react/components/alerts/ErrorAlert.tsx
sageknives: i would love to do it this way, I had it this way at one point. When you define them separately the hover state for functional components change to React.FC<ErrorAlertProps>
instead of React.FC<{props actually shown}>
The code completion works either way and you can always click on it and it shows you the d.ts file. What are your thoughts

packages/blocks-react/components/accordion/index.ts
```diff
@@ -0,0 +1 @@
+export { default as Accordian } from './AccordionSection';
```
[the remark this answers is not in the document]
sageknives: This was the only way I could figure out how to group files together to create the original require('blocks-react').Input structure. There is probably a way to do it all in the index.ts file but I timeboxed it and couldn't figure it out. We could create a separate ticket to try and figure this out.

packages/blocks-react/components/alerts/AlertBanner.tsx
```diff
@@ -73,4 +76,19 @@ AlertBanner.defaultProps = {
...
+  closable?: boolean,
+  icon?: React.ReactNode,
+  message?: string,
+  onClose?: () => void,
+  style?: React.CSSProperties | unknown,
+  title?: string,
+  /** other pass thru props */
+  [x: string]: any,
+};
+type AlertBannerState = {
```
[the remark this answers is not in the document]
sageknives: We don't need state but it gives blocks-ds developers code completion for props and state.

packages/blocks-react/components/buttons/LinkButton.tsx
```diff
@@ -13,16 +13,17 @@ const LinkButton = React.forwardRef((props, ref) => {
 
   if (useReactLink) {
     return (
+      //@ts-ignore
```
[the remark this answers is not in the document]
sageknives: For some reason it is saying react router link doesn't have a constructor, meaning it doesn't take any props, but it does so I probably should comment on this until it's figured out.

packages/blocks-react/components/accordion/index.ts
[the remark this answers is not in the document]
bleveque: oh I'm sorry, I just meant `Accordian` --> `Accordion` :) 👍

packages/blocks-react/components/alerts/ErrorAlert.tsx
bleveque: ah interesting, okay! Maybe having to click on the type and have the d.ts file open up wouldn't be so bad, but my gut feeling is that we should just show the whole thing on hover, as you have it, to make it as easy as possible. Maybe just worth documenting that in an eventual blocks typescripting guide? And we could always change it later if contributors find it confusing at all

packages/blocks-react/components/alerts/AlertBanner.tsx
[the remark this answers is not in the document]
bleveque: that could be nice, yeah, I think having a ticket to document that idea makes sense. just want to make sure we strike a balance between ease of consumption and ease of contribution -- maybe we could try to get some feedback from consumers before fully diving into documenting intrinsic types etc?

but either way, I think having some notes about what's expected of contributors would eventually be handy

packages/blocks-react/components/accordion/index.ts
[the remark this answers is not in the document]
sageknives: oh, yes will fix

bleveque:
hey @sageknives do you remember how to get code completion / hovering set up for files outside of blocks-react? I wanted to try out intellisense in a file like ToggleBasicExample.js, but I couldn't remember exactly how that was set up! Intellisense seems to work fine within blocks-react

packages/blocks-react/components/buttons/Button.tsx
```diff
@@ -1,25 +1,25 @@
-const React = require('react');
-const PropTypes = require('prop-types');
-const classnames = require('classnames');
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import classnames from 'classnames';
```
MagdalenaHenke: Screen Shot 2020-02-14 at 11 17 29 AM
 
I get sublime complain at me like this...although you didn't change anything about that package. Is that happening to you, too, @bleveque?

packages/blocks-react/dist/accordion/AccordionSection.d.ts
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react';
```
MagdalenaHenke: I think you might have talked about this but I can't remember - what are we putting these .d.ts files into `dist` for - is that so that other typescript projects using Blocks get intellisense? 

Also I feel like these should be minified.

packages/blocks-react/components/alerts/AlertBanner.tsx
```diff
@@ -1,11 +1,14 @@
-const React = require('react');
-const PropTypes = require('prop-types');
-const classnames = require('classnames');
+import * as React from 'react';
```
MagdalenaHenke: once this is rebased on master and prettier is a thing, you could run `npm run format:react` and `npm run format:docs` to just prettify all of these files in one swoop.

packages/blocks-react/components/alerts/AlertBanner.tsx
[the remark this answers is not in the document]
MagdalenaHenke: oooh yeah I'm really curious whether I'll loooooove this because code completion or mostly find it tedious to update 2 places :D

packages/blocks-react/components/alerts/ErrorAlert.tsx
MagdalenaHenke: Just reading this I'm ➕ ➕ for documentation...partially because not all blocks developers will be typescript experts (including me...yet) and documentation will be useful

packages/blocks-react/components/alerts/ErrorAlert.tsx
```diff
@@ -1,10 +1,11 @@
...
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
 
-const AlertBanner = require('./AlertBanner.jsx');
-const ErrorIcon = require('../icons/ErrorIcon.jsx');
+import AlertBanner from './AlertBanner';
+import ErrorIcon from '../icons/ErrorIcon';
 
-const ErrorAlert = props => (
+/** ErrorAlert */
```
MagdalenaHenke: I'm just curious why you added these comment titles

packages/blocks-react/components/buttons/index.ts
```diff
@@ -0,0 +1,5 @@
+export { default as ButtonDanger } from './ButtonDanger';
+export { default as ButtonDark } from './ButtonDark';
+export { default as ButtonPrimary } from './ButtonPrimary';
+export { default as ButtonSecondary } from './ButtonSecondary';
+export { default as LinkButton } from './LinkButton';
```
MagdalenaHenke: I'm honest, I haven't quite wrapped my head around the new export structure and why it's now what it is - I trust that this makes sense but I couldn't explain to other folks why it changed now. Maybe you'd be down to say words about this at me at some point?

packages/blocks-react/components/buttons/Button.tsx
bleveque: that is happening to me, too!

MagdalenaHenke: :D ugh whyyyyyyy. Sage, do you have an idea? 

The code seems to work all fine.

packages/blocks-react/package.json
```diff
@@ -17,8 +17,10 @@
   },
   "homepage": "https://github.com/floored/blocks#readme",
   "dependencies": {
-    "blocks-base": "^0.1.20",
-    "classnames": "^2.2.6"
+    "blocks-base": "^0.1.18",
```
MagdalenaHenke: I think we don't wanna go down in package number...

PR #206, opened 2020-01-23
Docs update: html language & document titles

MagdalenaHenke:
This is not a ticket, but I noticed these two quick (accessibility) wins while working on the docs, and they were too low hanging to not do.

1. Adding the `lang="en"` attribute to the main html element. That's necessary for screen readers, so they know which language to pronounce, but also helps other robots interpret the page correctly.
2. Making sure our document title changes on each site (each url should have a distinct document title). It's necessary for accessibility cause screen readers read out the new page title to alert users about where they moved to. Usually should match the h1 of the page, so that's how I implemented it. But it's also good for everyone: Now, if you have a ton of Blocks docs tabs open, the tab will tell you which one is where!


packages/blocks-docs/gatsby-config.js
```diff
@@ -50,6 +50,12 @@ module.exports = {
         icon: `${__dirname}/favicon.png`, // This path is relative to the root of the site.
       },
     },
+    {
+      resolve: "gatsby-plugin-html-attributes",
+      options: {
+        lang: "en",
+      },
+    },
```
MagdalenaHenke: > A Gatsby plugin to easily add HTML attributes like lang or xmlns.

https://www.gatsbyjs.org/packages/gatsby-plugin-html-attributes/

packages/blocks-docs/package-lock.json
```diff
@@ -2223,6 +2223,20 @@
       "resolved": "https://registry.npmjs.org/blob/-/blob-0.0.5.tgz",
       "integrity": "sha512-gaqbzQPqOoamawKg0LGVd7SzLgXS+JH61oWprSLH+P+abTczqJbhTR8CmJ2u9/bUYNmHTGJx/UEmn6doAvvuig=="
     },
+    "blocks-base": {
```
MagdalenaHenke: hmmm I npm installed the new gatsby package and then this mess showed up...should probably have another look at that

packages/blocks-docs/src/components/header.js
```diff
@@ -1,21 +1,19 @@
 import PropTypes from "prop-types"
 import React from "react"
+import Helmet from "react-helmet"
 
 const Header = ({ title, children }) => (
   <header>
+    <Helmet title={`${title}—Blocks by CBRE Build`} />
```
MagdalenaHenke: I'm making the title text contain the most specific information first, followed by the general, consistent "Blocks by CBRE Build". That's the best practice that I'm aware of.

packages/blocks-docs/src/components/header.js
```diff
@@ -1,21 +1,19 @@
...
     <div className="design-page-header">
-      <h1 className="blx-with-margin">
-        {title}
-      </h1>
-      <p className="blx-ui-text blx-with-margin">
-        {children}
-      </p>
+      <h1 className="blx-with-margin">{title}</h1>
```
MagdalenaHenke: sorry sublime prettiered this for me when I opened the file it's just whitespace changes

MagdalenaHenke:
@bleveque - maybe punting a tiny one your way?

packages/blocks-docs/src/components/header.js
```diff
@@ -1,21 +1,19 @@
...
+      <p className="blx-ui-text blx-with-margin">{children}</p>
     </div>
   </header>
-);
+)
 
 Header.propTypes = {
-  title: PropTypes.string.isRequired
-};
+  title: PropTypes.string.isRequired,
```
bleveque: does blocks have a trailing comma convention? I feel like I've seen it in a few places, but I forget if it's been mentioned as a convention!

packages/blocks-docs/package-lock.json
bleveque: fwiw it looks like running `lerna bootstrap` (maybe? I also ran `npm ci`...) got rid of these for me?

bleveque:
great! code looks good, and tests well
Screen Shot 2020-01-23 at 11 49 15 AM

Screen Shot 2020-01-23 at 11 49 02 AM

👍

packages/blocks-docs/src/components/header.js
MagdalenaHenke: Here's my hunch: 
blocks-docs has `prettier` installed (although I kind of hadn't realized that until a little bit ago. Many files aren't prettierd though.)
Emma even set up `npm run format` to format everything! Maybe I should just run that once on everything.

And the prettier setting on trailing commas [(from the .prettierrc file)](https://github.com/floored/blocks/blob/master/packages/blocks-docs/.prettierrc#L6) right now is es5, which means
> "es5" - Trailing commas where valid in ES5 (objects, arrays, etc.)

(from here: https://prettier.io/docs/en/options.html#trailing-commas)

So probably, where we don't do it is because people didn't and then they also didn't run prettier. I kind of like trailling commas, but I'm down to change the prettier setting if other people don't.

packages/blocks-docs/package-lock.json
MagdalenaHenke: @bleveque - no combination of `lerna bootstrap` and `npm ci` makes these go away from me - I'm starting to think these changes _should_ be here? Since I'm adding a new package, I gotta do `npm install` once (like I can't get away with `npm ci`), and no matter how often and when I `lerna bootstrap` or even `npm ci - for good measure` after that, these package-lock changes still happen. 
But `blocks-react` _also_ references block-base in its package-lock (https://github.com/floored/blocks/blob/master/packages/blocks-react/package-lock.json#L1049), and in my mind it makes sense that they would? 

Do you think they shouldn't be?

bleveque: hm yeah, it does make sense that they'd show up in package-lock, since blocks-docs has them both as dependencies.... Not sure why they were disappearing for me:

Screen Shot 2020-01-23 at 3 54 24 PM

if I `npm install` within packages/blocks-docs/, they do reappear! I may have cut off an `npm ci` partway through or something? not totally sure!

MagdalenaHenke: :D Maybe I'll get a tie breaker opinion. @tariqporter - do you think this change is ok or to be battled? Ben and I both seem a little "it's probably fine but maybe not."

MagdalenaHenke: I'm just going to cross my fingers and hope this is fine.

PR #207, opened 2020-01-24
Prettier - install in blocks-react and run everywhere.

MagdalenaHenke:
[Clubhouse](47117)

**Done in this PR:**
1. Installing [Prettier](https://prettier.io/) in blocks-react (it was already installed in blocks-docs)
2. running prettier on everything
3. setting up automatic prettifying of all staged files in our pre-commit hook

**For you to do:**
1. So make sure you set up your pre-commit hooks! `./bin/link_hooks`
2. Setup the prettier package in Sublime or VS Code or wherever and if you can bear it, run prettier on every save ;)

**Stuff that still needs doing in a future PR:**
1. Setup a prettify step in blocks-react happening on `gulp build` in case someone works on Blocks and doesn't set up their git hooks. 
2. It's not related immediately, but I wonder how eslint and prettier will interact...we just gotta enable eslint.


packages/blocks-react/package.json
```diff
@@ -4,7 +4,8 @@
   "description": "cbre build design system react components",
   "main": "index.js",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "format": "prettier --write components/**/*.{js,jsx}"
```
MagdalenaHenke: added this based on what Emma had put into `blocks-base` -  just to maki it easy to run prettier on everything.

packages/blocks-docs/.prettierrc
```diff
@@ -1,7 +1,7 @@
 {
   "endOfLine": "lf",
-  "semi": false,
+  "semi": true,
   "singleQuote": false,
   "tabWidth": 2,
-  "trailingComma": "es5"
+  "trailingComma": "none"
 }
```
MagdalenaHenke: These can be debated!

packages/blocks-react/package.json
MagdalenaHenke: Also figured I'd only run it on components and not on like...our gulp file, but maybe you'd actually prefer running it on everything?

packages/blocks-docs/.prettierrc
tariqporter: I think you can just remove the line `semi` as `true` is the default

MagdalenaHenke: I personally like it being explicit because then it's so easy to see the options, but if you feel strongly I can take it out.

tariqporter: I'm for keeping `trailingComma`. I've seen a general trend of packages having that as the default. ie from react's repo: https://github.com/facebook/react/blob/master/.prettierrc.js

bleveque: @tariqporter just out of curiosity, do you like trailing commas for the diffs? I don't have a super strong opinion on them, so I'm interested what people like or strongly dislike about them!

tariqporter: @bleveque Sorry for slow reply! Yes, I like that the diffs are cleaner, as well as it being slightly easier to add/remove things from an array/obj

hooks/pre-commit
```diff
@@ -1,11 +1,21 @@
...
 #
 # Currently not very smart - will not catch cases where you added
 # some generated files but missed others.
 
+
+# Prettify all staged js and jsx files not in dist
+echo "Prettify staged files"
+filesToPrettify=$(git diff --cached --name-only --diff-filter=ACMR "*.js" "*.jsx" | grep -v /dist/)
+echo "$filesToPrettify" | xargs ./node_modules/.bin/prettier --write --list-different
+echo "$filesToPrettify" | xargs git add
```
MagdalenaHenke: First actual code change - in the pre-commit hook, going through all the staged, non /dist/  js and jsx files and running them through prettier.

Based mostly on this: https://prettier.io/docs/en/precommit.html#option-5-bash-script
except adding the /dist/ grep to exclude files in dist, and adding `--list-different` to only write the files that got changed through prettier.

packages/blocks-docs/.prettierrc
```diff
@@ -1,6 +1,6 @@
 {
   "endOfLine": "lf",
-  "semi": false,
+  "semi": true,
   "singleQuote": false,
   "tabWidth": 2,
   "trailingComma": "es5"
```
MagdalenaHenke: so sounds like it's sort of a tie with a mild leaning towards trailing commas, so here we go

packages/blocks-docs/src/code_examples/ButtonBasicExample.js
```diff
@@ -1,29 +1,18 @@
...
-      onClick={() => {}}
-    />
-    <ButtonDark
-      text="Dark Button"
-      onClick={() => {}}
-    />
+    <ButtonPrimary text="Primary Button" onClick={() => {}} />
+    <ButtonSecondary text="Secondary Button" onClick={() => {}} />
+    <ButtonDanger text="Danger Button" onClick={() => {}} />
+    <ButtonDark text="Dark Button" onClick={() => {}} />
```
MagdalenaHenke: whether it gets spread or not is based on line length

packages/blocks-react/.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
+  "endOfLine": "lf",
+  "semi": true,
+  "singleQuote": false,
+  "tabWidth": 2,
+  "trailingComma": "es5"
+}
```
MagdalenaHenke: Added prettier and a `.prettierrc` also to blocks-react since right now it's a separate module.

packages/blocks-react/components/lists/SortableList.jsx
```diff
@@ -14,9 +14,10 @@ function isBefore(a, b) {
 
 function addOrRemoveClass(element, className, add) {
   if (add) {
-    if (element.className.indexOf(className) < 0) element.className += ` ${className}`;
+    if (element.className.indexOf(className) < 0)
+      element.className += ` ${className}`;
```
MagdalenaHenke: :D do I love this? no (new line in if statements but no curly brackets). But now it's the prettier-god's fault, so I won't complain.

packages/blocks-react/package.json
```diff
@@ -39,6 +40,7 @@
     "gulp-sourcemaps": "^2.6.5",
     "gulp-string-replace": "^1.1.2",
     "gulp-uglify": "^3.0.2",
+    "prettier": "^1.19.1",
```
MagdalenaHenke: installing prettier

packages/blocks-react/components/lists/SortableList.jsx
tariqporter: Oh, that's something that came up in site iq's prettier review. We manually added brackets around 1-line ifs that were wrapped by prettier

hooks/pre-commit
MagdalenaHenke: @bleveque - curious what you think!

MagdalenaHenke:
Ok this keeps failing on me deploy wise and I'll look into it but @tariqporter @sageknives and @bleveque (and @noreenwu if you'd like!) speak up now about prettier preferences or live with these forever.

MagdalenaHenke:
Hallelujah it works. 

Who'd like to have a look at this before it's merged?

MagdalenaHenke:
Opened a separate PR (yay small PRs) for a prettify gulp step.

tariqporter:
Nice 👍

MagdalenaHenke:
https://app.clubhouse.io/blocks/story/47117/setup-prettier-and-automatic-prettifying

hooks/pre-commit
```diff
@@ -1,11 +1,21 @@
...
+# 1. all staged files got prettified
+# 2. generated files have been added if React component files were changed
 #
 # Currently not very smart - will not catch cases where you added
 # some generated files but missed others.
 
+
+# Prettify all staged js and jsx files not in dist
+echo "Prettify staged files"
+filesToPrettify=$(git diff --cached --name-only --diff-filter=ACMR "*.js" "*.jsx" | grep -v /dist/)
```
bleveque: if you run `git commit -a`, do unstaged files get staged before this hook runs?

would we want to add .tsx once the typescript pr goes in, too?

hooks/pre-commit
```diff
@@ -1,11 +1,21 @@
...
+echo "$filesToPrettify" | xargs ./node_modules/.bin/prettier --write --list-different
```
bleveque: cool! I don't think I've used `xargs` before 🆒 

would `... | xargs npx run prettier --write ...` also work here? Not saying it's better, but I _think_ `npx` lets you run stuff from the project's .bin/

hooks/pre-commit
bleveque: cool! left a couple small comments above -- basically just want to make sure that all files you're intending to commit (including unstaged files if you use `git commit -a`, as I usually do!) get prettier-ed. This is great!

packages/blocks-docs/src/code_examples/ButtonBasicExample.js
bleveque: cool beans!

hooks/pre-commit
MagdalenaHenke: `npx prettier` it is, that works well and I think is better. Now I'm installing prettier on the top level, blocks-react and blocks-base...I think that's fine (I hadn't installed it top level before so now I'm confused about how this ever worked for me)

hooks/pre-commit
MagdalenaHenke: I tried this, and looks like it runs as we'd like it to for `git commit -a`. Added typescript extension!

MagdalenaHenke:
@bleveque - got back to your comments - thank you for those! Would you just give it a test, too?

hooks/pre-commit
bleveque: strange! yeah, and once we do that package restructuring, the duplicate installations should disappear, too, right?

bleveque:
Great! Only thing I noticed in testing is probably out of scope here: in the [tooltip demo](http://0.0.0.0:8080/components/tooltips), the tooltips seem to share state or something -- if you click from one `?` button to the next, the tooltips toggle rather than each new tooltip always showing up.

👍

hooks/pre-commit
bleveque: cool, thanks for testing that!

hooks/pre-commit
MagdalenaHenke: in my head, yeah, we'll just have one package.json / node_modules for everything then..._I think :D_

MagdalenaHenke:
Yeah, your tooltips are more like...toggle tips at this point. I also find it not super intuitive...although it mostly shows up really as noticeable because we have several of them next to each other. But that's also on production, I didn't break that  💃

bleveque:
true! in actual use cases, probably not much of a problem 👍

hooks/pre-commit
MagdalenaHenke: actually....now wondering whether I can't already get rid of the installations in the `packages` folders and just have it at the top folder...I had sort of assumed without thinking I needed it in the packages as well cause I'm using it in there...but I assume npm is smart enough to go looking further up in the folder tree to find dependencies...brb testing that

package.json
```diff
@@ -6,13 +6,16 @@
...
   },
   "scripts": {
     "build:base": "cd packages/blocks-base; gulp build",
     "build:docs": "cd packages/blocks-docs; gatsby build",
     "build:react": "cd packages/blocks-react; gulp build",
     "dev:docs": "cd packages/blocks-docs; gatsby develop -H 0.0.0.0 -p 8080",
-    "serve:docs": "cd packages/blocks-docs; gatsby build; gatsby serve"
+    "serve:docs": "cd packages/blocks-docs; gatsby build; gatsby serve",
+    "format:react": "cd packages/blocks-react; prettier --write components/**/*.{js,jsx,tsx}",
+    "format:docs": "cd packages/blocks-docs; prettier --write src/**/*.{js,jsx,tsx}"
```
MagdalenaHenke: @bleveque - followed your suggestion, just moved it all into the top folder. That's now the only one that's installing prettier!

(Also on this PR, made it so this one commit is the sum of all actual prettier implementation changes, and the second commit is prettifying everything - I felt that easier to review.)

bleveque: oh awesome, thanks!

bleveque:
looked at the first commit, and the code all looks good to me! I also ran `npm run format:react` and `npm run format:docs`, and both seemed to run through just fine 👍

PR #211, opened 2020-01-29
React Dropdown and Popover components keys

noreenwu:
React Dropdown and Popover components will use option.key || option.value to create unique key instead of utilizing option.text. The html needs to supply either or both key and value, thus the html code examples were also updated. 

https://app.clubhouse.io/blocks/story/46470/popover-menu-dropdown-item-s-key-shouldnt-be-tied-to-text


noreenwu:
@MagdalenaHenke @hwigmore : Hey, this is in response to the Clubhouse ticket . I saw same issue for Dropdowns as Popover so I looked at both. Thanks so much for reviewing!

packages/blocks-docs/src/code_examples/PopoverBasicExample.js
```diff
@@ -2,26 +2,33 @@ import React from "react"
 
 const { PopoverMenu } = require('blocks-react').Dropdowns;
 
+// Each menu item needs to include a value for either key or value; React component
+//   uses these two values to create a unique key for the menu item
 const exampleOptions = [
   {
     text: 'Action',
+    value: 'action',
```
hwigmore: ooh i totally forgot that this would apply to popovers as well as dropdowns! good find. but popover options never have values, i think (only text and onClick? @MagdalenaHenke is that right?) so maybe instead of forcing the user the pass a key in those cases, we should just make the key `option.key || option.value || option.text` because we know these will always at least have text?

i havent looked closely at blocks in awhile though so i would love lena's input too!

MagdalenaHenke:
https://app.clubhouse.io/blocks/story/46470/popover-menu-dropdown-item-s-key-shouldnt-be-tied-to-text

packages/blocks-docs/src/code_examples/PopoverBasicExample.js
MagdalenaHenke: I'm with you! I think we can handle Dropdowns and Popovers separately, though, and for Popovers have `key = option.key || option.text` and for Dropdowns be `key || value`.

packages/blocks-docs/src/code_examples/PopoverFlipExample.js
```diff
@@ -3,37 +3,46 @@ import React from "react"
 const { PopoverMenu } = require('blocks-react').Dropdowns;
 const { DownIcon } = require('blocks-react').Icons;
 
+// Each menu item needs to include a value for either key or value; React component
+//   uses these two values to create a unique key for the menu item
 const exampleOptions = [
   {
     text: 'Action',
+    value: 'action',
```
MagdalenaHenke: yeah, I'd prefer to not force people to submit a key or value for popovers when it isn't needed. I like an approach similar to what you did for Dropdowns - except here using `text` as the key _unless_ a specific `key` was passed in and then using that.

noreenwu:
@MagdalenaHenke @hwigmore  Thanks for your thoughtful comments. I changed PopoverMenu to utilize 'option.key' if provided and otherwise 'option.text'.  As a result, the example component changed as well. Would you take another look? Suggestions welcome!

packages/blocks-docs/src/code_examples/PopoverFlipExample.js
```diff
@@ -3,37 +3,46 @@ import React from "react"
...
 
 const exampleOptions2 = [
   {
     text: 'Correct Action',
+    value: 'correctaction',
     onClick: () => {}
   },
   {
     text: 'Another Action',
+    value: 'anotheraction',
```
MagdalenaHenke: In my head, we don't need these `value` properties anymore

packages/blocks-docs/src/code_examples/PopoverIconExample.js
```diff
@@ -3,26 +3,38 @@ import React from "react"
 const { PopoverMenu } = require('blocks-react').Dropdowns;
 const { MoreVerticalIcon, MoreIcon } = require('blocks-react').Icons;
 
+// Each menu item needs to include a value for either key or value; React component
+//   uses these two values to create a unique key for the menu item
 const exampleOptions = [
   {
     text: 'Action',
+    value: 'action',
```
MagdalenaHenke: same here, `Popover` doesn't use `value` anymore now again, doesn't it?

hwigmore:
nice! i tested this locally and it got rid of the errors i was seeing. besides lena's comments 👍 from me!!

packages/blocks-docs/src/code_examples/PopoverIconExample.js
noreenwu: Yes, thanks! I have removed the value field from the 2nd and 3rd examples (PopoverFlip and PopoverIcon) and put in a duplicate menu item example using key. What do you think?

MagdalenaHenke:
Looks good to me! 🌮 

(I use 👍 and 🌮interchangeably for no good reason at all.)

noreenwu:
@hwigmore @MagdalenaHenke  Thank you both for the review and comments! While working on the documentation reorg, it occurred to me that what I put in the React comments for optionally adding a "key" value should perhaps be listed in the section titled Examples and Guidelines. What do you think? It only applies to React and not the plain HTML. If you think something should be said, I will add those changes to this same branch; otherwise we can go ahead and merge this with master.  Thanks, again!

MagdalenaHenke:
For posterity: we talked in person and we think it's fine to have the comments in code for now! Maybe in the future, after the documentation reorganization can live in the "code" tab, but good for now!

PR #214, opened 2020-01-31
add eslint static analysis

MagdalenaHenke:
Adding `eslint`. 

So Emma had originally not enabled this, because she prefers to not have the linter scream at her as she codes. Fair enough. But it sounded like in this bunch, people would prefer having the linter enabled all the time (especially in prep for cross-team collaborations), and one can just manually disable the linter locally if you don't like it.

I went through a bunch of our files to make sure to disable some warnings we don't want. But I didn't fix an of the remaining warnings - I figured that'd just make a bunch of conflicts again, and instead people can clean up as they edit files.

Now the main warning that'll remain all over the place is about preferring to spread props (instead of doing `props.blabla`). It's annoying to see everywhere, but we also _want_ to make that change so felt like I shouldn't disable that warning.
I also added the a11y linter cause I like it.


.eslintrc.json
```diff
@@ -0,0 +1,22 @@
+{
+  "extends": ["airbnb", "prettier"],
```
MagdalenaHenke: Extending `prettier` to disable all the rules that conflict with prettier.
https://prettier.io/docs/en/integrating-with-linters.html#disable-formatting-rules

.eslintrc.json
```diff
@@ -0,0 +1,22 @@
...
+  "rules": {
+    "func-names": "off",
+    "import/no-extraneous-dependencies": "warn",
+    "no-console": "off",
+    "no-param-reassign": "off",
+    "no-plusplus": "off",
```
MagdalenaHenke: the linter is for `i += 1` over `i++`. On Spacer we do this, and I still think `i++` is such a convention to see everywhere, and the issues around it so rarely come up, that I disabled that warning...unless any of you actually would prefer us to switch to `+=1` type stuff.

.eslintrc.json
```diff
@@ -0,0 +1,22 @@
...
+
+     "react/jsx-props-no-spreading": "off",
+     "react/destructuring-assignment": "warn",
+     "react/jsx-fragments": "off",
+
+    "jsx-a11y/label-has-for": "off",
```
MagdalenaHenke: disabling this one because it doesn't take into account implicitly linking labels and their input by wrapping the input in a `<label` element.

MagdalenaHenke:
https://app.clubhouse.io/blocks/story/47233/setup-eslint

package.json
```diff
@@ -6,6 +6,12 @@
     "npm": "6.8.0"
   },
   "devDependencies": {
+    "eslint": "^6.8.0",
+    "eslint-config-airbnb": "^18.0.1",
+    "eslint-config-prettier": "^6.10.0",
+    "eslint-plugin-import": "^2.20.0",
+    "eslint-plugin-jsx-a11y": "^6.2.3",
+    "eslint-plugin-react": "^7.18.0",
```
MagdalenaHenke: I installed eslint in the top level folder (blocks, _above_ the `packages` ones), because it's mainly used during development to give you warnings and you'll always be working in the entire repo for that.

For prettier, I installed it separately in the blocks-base and block-react packages because I didn't think about it too deeply since Emma had already installed it in _one_ of those two, so I just added it to the other one. 
Is that cool with you all?

MagdalenaHenke:
@tariqporter or @sageknives ... do we love it, do we hate it?

.eslintrc.json
```diff
@@ -0,0 +1,23 @@
...
+    "func-names": "off",
+    "import/no-extraneous-dependencies": "warn",
+    "no-console": "off",
+    "no-param-reassign": "off",
+    "no-plusplus": "off",
+
+     "react/jsx-props-no-spreading": "off",
+     "react/destructuring-assignment": "warn",
+     "react/jsx-fragments": "off",
+     "react/jsx-filename-extension": "off",
```
MagdalenaHenke: this one warns you if you have jsx in files that are not .jsx files. Which we have a ton, like in [`colors.js`](https://github.com/floored/blocks/blob/master/packages/blocks-docs/src/pages/styles/colors.js). 

I'm not sure _why_ we have that as a js file and not a jsx one, though

MagdalenaHenke:
@bleveque @sageknives - adding eslint for static analysis. I realized I never tagged anyone on this. Would you have a look?

.eslintrc.json
bleveque: fwiw I like `i++`!

.eslintrc.json
bleveque: prob irrelevant, since this is disabled, but out of curiosity, do you know if it would complain if there were jsx in a .tsx file?

MagdalenaHenke: I think it _would_ complain, according to the internet! We could also specify specifically which extensions are ok with something like this:
```json
"react/jsx-filename-extension": [1, {
      "extensions": [".jsx", ".js", ".tsx"]
 }],
```

But I figured we might disable it altogether.

bleveque:
cool! 👍

PR #220, opened 2020-02-06
gitignore DS_Store

MagdalenaHenke:
@sageknives - you pointed this out in your typescript demo and yeah...should probably just gitignore DS_Store.

I made sure it works for all sub folders as well.


MagdalenaHenke:
Or maybe @noreenwu I'll throw this one your way

noreenwu:
Looks good to me!

PR #224, opened 2020-03-09
moved Modals and Tabs under Components, and removed Patterns subsection

noreenwu:
@MagdalenaHenke  Modals and Tabs are now included as Components rather than being in their own Patterns grouping


noreenwu:
https://app.clubhouse.io/blocks/story/48051/modals-and-tabs-should-be-components-not-patterns

MagdalenaHenke:
🌮 from me.

Part of me thought "you know what, if we're already reshuffling, we might as well reshuffle the other stuff to fit along with Runi's ['new' outline]" (which 
  includes getting rid of the "Help" section and renaming/reshuffling things), buuuuut I think it might be better / easier rebasing wise to first get the mdx PR in and _then_ do those reshuffles, since they'll affect each other.

MagdalenaHenke:
So then if we wait with the other reshuffles, would you ticket them and maybe reference the excel spreadsheet so we remember they're "todo" still? https://docs.google.com/spreadsheets/d/1GaOWY-q9eNH0NUazDC3VX-izO_QmAC0j6rJmZPRo3fg/edit#gid=0

noreenwu:
Got it. Created a ticket referencing the spreadsheet (I will have to come in to review the mdx PR because I can't get it to work on this machine.)

PR #225, opened 2020-03-12
Hooksify HOCs

MagdalenaHenke:
@sageknives - I just kind of ran with the "buuut what if we hooksified HOCs?" comment because I love turning things into hooks and might as well solve Typescript weirdness sooner.

I built this one on top of the `eslint` branch, because I wanted to get the eslint-hooks warnings. 

I think there's still a little buggy on this one, but the overall holds up, I think?

The one downside is that if we publish this, we're forcing everyone who _wants to use these wrapper functionalities on their own components_ to be on a version of React that includes hooks. I'm actually ok with that. I know in Spacer we have one example of using `closeOnClick` in one of our own components...and oh well, it'll just have to be refactored. How do you feel about that?


.eslintrc.json
```diff
@@ -6,12 +6,12 @@
     "no-console": "off",
     "no-param-reassign": "off",
     "no-plusplus": "off",
```
MagdalenaHenke: look at this PR hiding whitespace changes!

.eslintrc.json
```diff
@@ -6,12 +6,12 @@
...
-     "react/destructuring-assignment": "warn",
-     "react/jsx-fragments": "off",
-     "react/jsx-filename-extension": "off",
-
+    "react/jsx-props-no-spreading": "off",
+    "react/destructuring-assignment": "warn",
+    "react/jsx-fragments": "off",
+    "react/jsx-filename-extension": "off",
+    "react-hooks/rules-of-hooks": "error",
+    "react-hooks/exhaustive-deps": "warn",
```
MagdalenaHenke: adding hooks eslint rules

packages/blocks-react/components/dropdowns/Dropdown.jsx
```diff
@@ -4,21 +4,54 @@ const classnames = require("classnames");
...
+    style,
+    className,
+    label,
+    autoFocus,
+    icon,
+    options,
+    scrollable,
+    onSelect,
+    invalidErrorMessage,
+  } = props;
```
MagdalenaHenke: I did this because one of the decisions was to move towards destructuring props at the top of files, and now eslint also complains about it, so I figured I might as well.

packages/blocks-react/components/dropdowns/Dropdown.jsx
```diff
@@ -4,21 +4,54 @@ const classnames = require("classnames");
...
+    "blx-disabled": disabled,
+    "blx-invalid": !isValid,
   });
 
-  if (props.value) {
-    for (let i = 0; i < props.options.length; i++) {
-      const option = props.options[i];
-      if (option.value === props.value) {
+  if (value) {
+    for (let i = 0; i < options.length; i++) {
```
MagdalenaHenke: so most of these changes are just getting rid of `props`

packages/blocks-react/components/dropdowns/Dropdown.jsx
```diff
@@ -65,44 +95,41 @@ const DropdownMenu = props => {
...
+      {!isValid && (
+        <span className="blx-invalid-input-message">{invalidErrorMessage}</span>
       )}
     </div>
   );
 };
 
-DropdownMenu.propTypes = {
+Dropdown.propTypes = {
+  afterDropdownClose: PropTypes.func,
```
MagdalenaHenke: eslint helpfully called this out for me

packages/blocks-react/components/dropdowns/Dropdown.jsx
```diff
@@ -125,24 +152,23 @@ DropdownMenu.propTypes = {
...
-  triggerContent: null,
+  disabled: false,
   onSelect: () => {},
   autoFocus: false,
   isValid: true,
   invalidErrorMessage: null,
 };
 
-module.exports = closeOnClick(keyControlledMenu(DropdownMenu));
+module.exports = Dropdown;
```
MagdalenaHenke: moved into hooks - upside is that they now _could_ also share state not just from closeOnClick to keyControlledMenu (which happens right now via props), but also the other way around!

packages/blocks-react/components/dropdowns/PopoverMenu.jsx
```diff
@@ -62,13 +85,10 @@ const PopoverMenu = props => {
 PopoverMenu.propTypes = {
   className: PropTypes.string,
   style: PropTypes.object,
-  isOpen: PropTypes.bool,
```
MagdalenaHenke: removed all the ones eslint called out as unused

packages/blocks-react/components/wrappers/keyControlledMenu.jsx
```diff
@@ -1,99 +1,28 @@
...
-      this.onKeyUp = this.onKeyUp.bind(this);
-      this.onTriggerFocus = this.onTriggerFocus.bind(this);
+const { createRef, useState, useEffect } = require("react");
+
+function useKeyControlledMenu({ options, isOpen, open, close, onChange }) {
+  const [onFocusIdx, setOnFocusIdx] = useState(-1);
+  const [optionsRefs, setOptionsRefs] = useState(() => {
+    const newOptionsRefs = [];
+    for (let i = 0; i < options.length; i++) {
+      newOptionsRefs[i] = createRef();
```
MagdalenaHenke: I feel a little iffy about this - couldn't do `useRef` because then I got complains about not doing `useRef` in a loop and how hooks always need to be called in the same order, which wouldn't be the case if options.length changes. `createRef` seems to still do what I want, so I went for the non-hooks version of ref creation - with mixed feelings.

MagdalenaHenke:
@Jinsoo94 - this isn't 100% bug free, but if you have time, I'd love to hear your thoughts. I know you were as excited as I am about turning these HOCs into hooks.

@sageknives - I know this was also on your agenda to simplify typescript stuff, how do you feel about it? 

Anyone: for hooks to work you might have to follow the README instructions from this PR: https://github.com/floored/blocks/pull/179

packages/blocks-react/components/dropdowns/Dropdown.jsx
```diff
@@ -4,21 +4,54 @@ const classnames = require("classnames");
```
MagdalenaHenke: scroll all the way down to `closeOnClick` and `keyControlledMenu` to see the changes to those files, turning them into shared hooks. The other files are the ones _consuming_ either one of them, so I had to change those.

packages/blocks-react/components/dropdowns/Dropdown.jsx
```diff
@@ -4,21 +4,54 @@ const classnames = require("classnames");
...
+    className,
+    label,
+    autoFocus,
+    icon,
+    options,
+    scrollable,
+    onSelect,
+    invalidErrorMessage,
+  } = props;
+  const triggerRef = React.useRef(null);
```
MagdalenaHenke: Now, instead of closeOnClick creating this ref and passing it down, I decided to pass it _into_ the hook. Although I guess I could also create it in the hook, and return it from there.

packages/blocks-react/components/wrappers/closeOnClick.jsx
```diff
@@ -1,88 +1,38 @@
...
+const { useState, useEffect } = require("react");
 
-function closeOnClick(WrappedComponent) {
-  class ComponentWithCloseOnClick extends React.Component {
-    constructor(props) {
-      super(props);
+function useCloseOnClick(ref, afterCloseOnClick) {
+  const [isOpen, setIsOpen] = useState(false);
 
-      this.state = { isOpen: !!props.isOpen };
```
sageknives: so this HOC received a prop for isOpen, is that no longer required for the hook version?

sageknives: and is so then the components that use this need to be able to accept it as a prop

